PR #34, opened 2019-01-31
Fetching schema by url in library

### LINKED ISSUE #32: Add possibility to pass schema by url
**Description**

At the moment, we have only a possibility to pass AsyncApi specification by string (yaml) or by json. We good improvement would be add possibility to pass schema by url and then fetching content by our logic. Of course, we have a already this logic, but separate in playground app.

magicmatatjahu:
**Description**

- Add feature to fetch schema from external resources in library package

**Related issue(s)**
Resolves #32


README.md
```diff
@@ -45,6 +49,8 @@ The list of props for the AsyncApi React component includes:
 
      > **NOTE:** The `Partial<T>` type means that every field in the `T` type is optional.
 
+   > **NOTE:** You must pass `schema` or `urlSchema` to component for correct work of component.
```
fmvilas: What about rewriting this copy?

Either `schema` or `urlSchema` prop must be present.

library/src/containers/AsyncApi/AsyncApi.tsx
```diff
@@ -98,6 +121,8 @@ class AsyncApiComponent extends Component<AsyncApiProps, AsyncApiState> {
 
     if (!(validatedSchema && validated)) return null;
 
+    console.log(validatedSchema);
```
fmvilas: Was it forgotten here?

library/src/containers/AsyncApi/AsyncApi.tsx
```diff
@@ -53,12 +60,28 @@ class AsyncApiComponent extends Component<AsyncApiProps, AsyncApiState> {
   }
 
   async componentWillMount() {
```
fmvilas: Heads up `componentWillMount` is deprecated. Check out: https://reactjs.org/blog/2018/03/27/update-on-async-rendering.html.

library/src/containers/AsyncApi/AsyncApi.tsx
```diff
@@ -16,7 +22,8 @@ import ErrorComponent from '../Error/Error';
 import { AsyncApiWrapper } from './styled';
 
 export interface AsyncApiProps {
-  schema: string | Object;
+  schema?: string | Object;
+  urlSchema?: FetchingSchemaInterface;
```
fmvilas: Can we enforce that either schema or urlSchema must be provided? I know it's possible with prop-types, but not sure about Typescript.

library/src/containers/AsyncApi/AsyncApi.tsx
magicmatatjahu: Hah :D I always forget about console.log in my code. Thanks!

library/src/containers/AsyncApi/AsyncApi.tsx
fmvilas: I didn't explain myself correctly. I mean having both schema and urlSchema, like it's right now, but enforcing in the code the rule of having just one of them.

magicmatatjahu: I implemented my proposition from second comment in this thread. Please see: https://github.com/asyncapi/asyncapi-react/pull/34/files#diff-2fa873daa631d9246738398633704fd9R25

fmvilas: Nice! I love this solution. Just one question: I see it says `string | Object | FetchSchemaInterface`, will it always fallback into `Object`? Saying because an object of the form FetchSchemaInterface is also an object, right? Or will typescript take care of this?

magicmatatjahu: Thanks :) You can pass schema as string (json or yaml), or Object (this is reference to AsyncAPIInterface) or FetchSchemaInterface. I agree with you that FetchSchemaInterface is also a Object, but Object with appropriate fields, so if dev pass Object as FetchSchemaInterface to component then component will execute whole code as FetchSchemaInterface, if pass as Object, then code validate this Object as AsyncApi and show errors if this spec is invalid :) You have to look at it in terms of static typing, not dynamic.

fmvilas: Understood. Thanks! :)

README.md
```diff
@@ -28,9 +28,9 @@ Check out this simple sandbox application that uses the asyncapi-react component
 
 The list of props for the AsyncApi React component includes:
 
-   - **schema: string | AsyncApiInterface**
+   - **schema: string | AsyncApiInterface | FetchingSchemaInterface**
 
-     The `schema` property is required and contains AsyncAPI specification. It should be one of the `string` or [`AsyncApiInterface`](./library/src/types.ts#L13) type. For more information on what it contains and what it should look like, read [AsyncAPI Specification](https://github.com/asyncapi/asyncapi#asyncapi-specification).
+     The `schema` property is required and contains AsyncAPI specification. It should be one of the `string` or [`AsyncApiInterface`](./library/src/types.ts#L13) type or [`FetchingSchemaInterface`](./library/src/helpers/fetchSchema.ts#L1) object for fetching schema from external resource. For more information on what it contains and what it should look like, read [AsyncAPI Specification](https://github.com/asyncapi/asyncapi#asyncapi-specification).
```
kazydek:
```suggestion
     The `schema` property is required and contains AsyncAPI specification. Use the `string` type, the [`AsyncApiInterface`](./library/src/types.ts#L13) type, or the [`FetchingSchemaInterface`](./library/src/helpers/fetchSchema.ts#L1) object to fetch the schema from an external resource. For more information on what it contains and what it should look like, read [AsyncAPI Specification](https://github.com/asyncapi/asyncapi#asyncapi-specification).
```

PR #56, opened 2019-08-28
add unit tests

m00g3n:
**Description**
- add unit tests

**Related issue(s)**
kyma-project/console#684


library/package.json
```diff
@@ -50,7 +50,7 @@
     "postpack": "rm -rf ./README.md && rm -rf ./LICENSE"
   },
   "dependencies": {
-    "asyncapi-parser": "^0.7.0",
+    "asyncapi-parser": "^0.7.1",
```
magicmatatjahu: Already is `0.8.2` version.

library/src/helpers/parser.ts
```diff
@@ -12,25 +10,37 @@ import {
...
+) => Promise<AsyncApi>;
+
+type ParseDocumentFromURL = (
+  url: string,
+  requestOptions?: RequestInit,
+  parserOptions?: ParserOptions,
+) => Promise<AsyncApi>;
+
+export default class Parser {
+  private parseDocument: ParseDocument;
```
magicmatatjahu: `parse` and `parseFromURL`.

library/src/helpers/parser.ts
```diff
@@ -12,25 +10,37 @@ import {
   AsyncApiProps,
 } from '../types';
 
+import { UNSUPPORTED_SCHEMA_VERSION } from '../constants';
+
 type ParserOptions = AsyncApiProps['parserOptions'];
 
-import { UNSUPPORTED_SCHEMA_VERSION } from '../constants';
+type ParseDocument = (
```
magicmatatjahu: `ParseSchema` and `ParseSchemaFromURL`

library/src/helpers/parser.ts
```diff
@@ -68,10 +70,7 @@ class Parser {
       return { data: null, error: { message: err.message } };
     }
 
-    if (
-      err.parsedJSON &&
-      !this.isCorrectSchemaVersion(err.parsedJSON.asyncapi)
-    ) {
+    if (err.parsedJSON && err.parsedJSON.asyncapi.startsWith('1')) {
```
magicmatatjahu: Please revert this to `isCorrectSchemaVersion` function.

m00g3n: - `isCorrectSchemaVersion` only checks the first character e.g. `-1.2.3` is not a correct schema version but the function would return `true`
- it is oneliner used just in this place

library/src/helpers/parser.ts
m00g3n: This is the tricky one we have a class that already has `parse` method.
I see two solutions:

1. rename class method from `parse` to `do` than rename `parseDocument` to `parse` and `parseDoumentFromURL` to `parseURL`
1. rename from `parseDocument` to `parseFn` (or similar) and `parseDoumentFromUR`to `parseURLFn`

magicmatatjahu: 3. rename to `parseSchema{FromURL}`

Issue #67, opened 2019-09-13
library crashes when invalid json schema


rmelian:
**Description**

when writing a document that is not valid the library crashes

**Expected result**
should not crash. should show an error indicating what's invalid or ignore the error

**Actual result**
![image](https://user-images.githubusercontent.com/4565267/64896095-8fcdd280-d64c-11e9-9cae-a278c2d3c5d7.png)

**Steps to reproduce**

write the following example
```javascript
asyncapi: '2.0.0'
id: 'urn:myapi'
info:
  title: My API
  version: '1.0.0'
channels:
  mychannel:
    publish:
      message:
        $ref: '#/components/messages/testMessage'

components:
  messages:
    testMessage:
      payload:
        $ref: '#/components/schemas/testSchema'
  schemas:
    testSchema:
      type: object
      properties:
        name:
          type: string
    order:
      type: object
      properties:
        name:
          type: string
    activity:
      type: o
```

the Activity type is invalid.

**Troubleshooting**

This is the validatedSchema returned from AsyncApi.tsx
```javascript
const { validatedSchema, error } = this.state;
```
![image](https://user-images.githubusercontent.com/4565267/64896329-606b9580-d64d-11e9-9558-6773d51c2a86.png)

Error from Chrome console
```
Schema.js:47 Uncaught TypeError: Cannot create property 'description' on string 'o'
    at SchemaComponent (Schema.js:47)
    at renderWithHooks (react-dom.development.js:15821)
    at updateFunctionComponent (react-dom.development.js:17705)
    at beginWork$1 (react-dom.development.js:19371)
    at HTMLUnknownElement.callCallback (react-dom.development.js:363)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:412)
    at invokeGuardedCallback (react-dom.development.js:466)
    at beginWork$$1 (react-dom.development.js:24570)
    at performUnitOfWork (react-dom.development.js:23502)
    at workLoopSync (react-dom.development.js:23480)
    at renderRoot (react-dom.development.js:23155)
    at runRootCallback (react-dom.development.js:22809)
    at react-dom.development.js:11886
    at unstable_runWithPriority (scheduler.development.js:674)
    at runWithPriority$2 (react-dom.development.js:11834)
    at flushSyncCallbackQueueImpl (react-dom.development.js:11881)
    at flushSyncCallbackQueue (react-dom.development.js:11869)
    at scheduleUpdateOnFiber (react-dom.development.js:22667)
[...]

PR #68, opened 2019-09-13
type checking if schema property is not string

rmelian:
**Description**

- check if properties are objects and valid 
-  check for schema type  === 'object'

**Related issue(s)**

#67 #69


rmelian:
I'll update this PR, thanks

rmelian:
PR updated

Issue #75, opened 2019-10-30
Provide an Angular-Component


mischahaenen:
**Description**
For my company I integrate the Kafka API's into the Angular website. The goal is to be able to display and manage the APIs. I am currently integrating the React component into the existing Angular application.
I realized this with the contribution of Stackoverflow: https://stackoverflow.com/questions/47714221/using-react-component-library-in-an-angular-4-application. So the compatibility with Angular applications would be given.

**Reason**
Currently it is difficult to test or debug the React component. I can't derive errors well either. I think these bugs could be fixed if you offered an Angular component. Alternatively, web components are certainly possible. API management is not well known at the moment but I think that many developers will be happy if a component would be available.

derberg:
Hey, you can close but only if you are not interested in contributing Angular component together :) so the thing is, keeping 3 components up to date is hard, but we could together maybe come up with a good idea how to do it efficiently. Maybe some code can be abstracted to separate npm modules and reused in those different components. Maybe we could do something to make working on components for different frameworks much more pleasure and not a nightmare :D

Of course at the end it all depends, how hard it really is to maintain react component in angular app, is it so hard that it is better to work on separate Angular component.

mischahaenen:
Good idea :) my idea would be that you make a bundle out of the react component. the practical thing would be that you don't have to compile the component anymore and could integrate it into all frameworks as well as angular. Plus you only have to update one component. I have already integrated the SwaggerUIBundle (https://github.com/activelamp/swagger-ui-bundle) and it works fine in my application. It would be cool if you could publish it, so the bundle is always up to date. Embedding a React component is actually not a problem, but when compiling it comes to errors so I came up with the idea of creating a bundle. What do you think about it?

derberg:
tbh I've never met before with the solution you proposed, and the fact that the repo you linked is not actively maintained for years do not make me an optimist here :)

I checked with @aerfio and @magicmatatjahu and we have a `?` mark here. Is this known pattern, do you know some articles or other examples that follow the same approach?

derberg:
What about some recent patterns, like wrapping react in web component, or maybe this approach backed by Microsoft https://github.com/microsoft/angular-react ?

aerfio:
Some comments here from my side: same thing as @derberg said about that one repo, without much investigation I'd gravitate towards something like swagger-ui-**_dist_**, or maybe web components

[...]

PR #76, opened 2019-11-28
support webcomponent for asyncapi component

hesyar:
Here is a basic way to create a webcomponent out of the AsyncApi React Component. The code isn't perfect but is a proof of concept how to achieve to create a web component and use it in your angular (or other js framework/library) app. 

**Usage**

After installing the package with `npm install`. You'll find a bundled Javascript file called `~/asyncapi-react/library/lib/async-api-web-component.js`.

This can be included in your angular.json file (if you use angular-cli) like this: 
`
            "scripts": [
              "src/assets/async-api/async-api-web-component.js"
            ]
`

or just add it in your index.html file.

Then in any of your angular component you can use it as a normal component `<async-api-component [schema]="spec"></async-api-component>` where you could also pass a a config propery.

Make sure that your module as `schemas: [CUSTOM_ELEMENTS_SCHEMA]`set, otherwise you'll get an error in angular that is doesn't know the component.

**Related issue(s)**
Workaround for #75


magicmatatjahu:
@hesyar Thank for contribution and sorry for late response.

As you wrote, this is concept from your side, so I won't do review in classic way, but only refer to your solution. I don't have a problem with web-component in combining it with some library (React or Vue) or framework (Angular), but your workaround changes a little bit the main component of library, what I would like to avoid. Also importing styles inside component is probably dictated by 
hermeticity of styles in Angular (correct me if I'm wrong) - of course they can be overwritten by the `host` ng-selector in the component as I know, but I also would like to avoid it.

Instead of using web component, I would like to create something like a bundle https://github.com/swagger-api/swagger-ui, which is mentioned in related issue. I use it myself many times and it and problems only appeared with style overwriting.

I don't know how long it would take to write a bundle for other libraries (not only React), but if it takes a long time, we can do it according to your suggestion and then rewrite it to the bundle like [here](https://github.com/swagger-api/swagger-ui/blob/master/src/core/index.js). What do you think @aerfio @hesyar @mischahaenen @derberg

hesyar:
@magicmatatjahu you are totally right. I should not have put the styles inside the component but this was a quick and dirty solutions for me. Of course if you could find a solution to create a bundle like the swagger UI it would solve a lot of issues. I tried that also but as I've created the bundle, I could not use the component inside my angular app. The bundle and combination of the web component however worked properly. With this PR I wanted to demonstrate a working solutions. I was hoping that this approach might come in use for use.

magicmatatjahu:
@hesyar Ok, so if I find time today, I try to create a bundle distribution.

aerfio:
Hey good concept :D I think that if such a need arises that people want to have this react component working e.g in Angular, and we have no time currently to work on it, then we should go along with proposed plan

I think we were never especially attatched to that swagger-like-bundle

The only problem I have with adding something like this to repo means that's another thing to maintain, and webpack configs are fine if they are working properly, but if something doesn't work then it soo much fun with them 😢 

I have an idea - would it work (from a technical point of view) if you created separate repo, npm-installed it there and re-exported, but in such a way that is presented here? Because if someone is looking for Angular component s/he would not look into React component repo

magicmatatjahu:
@aerfio Always we can rename repo to `asyncapi-ui` :) Mono repo here is not a problem, we only must create a good structure of repo, with hermetically folders for pure react component and bundle, like in mentioned openapi.

aerfio:
But we would have to have such a config in a repo eventually anyway, doesn't matter what it is so it's just me thinking aloud :d 

as for the repo name we should ask @fmvilas for opinion here

hesyar:
A good solutions here would be to create a mono repo and then multiple folders for react, angular (maybe vue) and so on. The component itself will be developed in react itself. The other folders for angular and so on will be in best case just some config to make it usable in Angular. This way, there is no reason to maintain multiple repositories and the angular component and react component are always in sync.

derberg:
I’m also in favor of mono repo + rename to asyncapi-ui

PR #98, opened 2020-04-13
display enums in component if available

### LINKED ISSUE #93: Support displaying enums
**Description**

When enums are used within the spec file, they are not displayed. 
As we provide enums like following:
image

We would expected here enums to be shown like in the generator itself but rather get the following:
image

**Reasons**

We want to be able to see enums in the component

**Attachments**

hesyar:
**Description**

Displays enums when they are set. Style was taken from the asyncapi/generator

**Related issue(s)**
Resolved #93

### DIFF
--- a/library/src/containers/Schemas/SchemaProperties.tsx
+++ b/library/src/containers/Schemas/SchemaProperties.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import merge from 'merge';
 
+import { bemClasses } from '../../helpers';
 import { TypeWithKey, Schema } from '../../types';
 import {
   Markdown,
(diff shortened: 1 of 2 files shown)


derberg:
@hesyar thanks man! can you please rebase to latest master? I've added a new workflow for running tests on PRs

hesyar:
its rebased now

hesyar:
When will the review of this PR be done?

michal-hudy:
@magicmatatjahu @aerfio please take a look 😄

library/src/containers/Schemas/SchemaProperties.tsx
```diff
@@ -45,12 +46,32 @@ const schemaPropertiesAccessors: Array<TableAccessor<SchemaElement>> = [
...
-    el.schema.content.description && (
-      <Markdown>{el.schema.content.description}</Markdown>
-    ),
+  el => {
+    const enumElements = getEnumHTMLElements(el.schema);
+    return (
+      el.schema.content.description && (
+        <div>
+          <Markdown>{el.schema.content.description}</Markdown>
+          {enumElements.length && <div>Enum: {enumElements}</div>}
```
magicmatatjahu, line 55: I found a problem with conditional rendering in this place.

For example. If we have schema like that (this is entry of schemas field):

```yaml
    turnOnOffPayload:
      type: object
      properties:
        command:
          type: string
          enum:
            - on
            - off
          description: Whether to turn on or off the light.
        sentAt:
          $ref: "#/components/schemas/sentAt"
```

Then by your code, rendered enums look like:
image

According to React docs https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator conditional rendering in React only avoid false, null and undefined values. The other values (also 0 - second row) are render - if condition return false. So your code for return enums should looks like:

```js
{enumElements.length > 0 && <div>Enum: {enumElements}</div>}
```

or

```js
{enumElements.length ? <div>Enum: {enumElements}</div> : null}
```

I also wonder if the description place is good for rendering enums. Better would be default, but since you had to wait a long time for review, we can leave it.

hesyar: Hej @magicmatatjahu no worries. Thanks a lot. I changed the code like you mentioned above. Thanks for the input ;) I hope I didn't make any pressure :D

derberg:
@magicmatatjahu remember about conventional commits when you will squash and merge 😄 so the release will be automatic

derberg:
@all-contributors please add @hesyar for bug, code

PR #133, opened 2020-09-01
feat: Add more property labels and render complex objects

fmvilas:
**Description**

- Add more property labels: minimum, maximum, exclusiveMinimum, exclusiveMaximum, uniqueItems, minLength, maxLength, minItems, maxItems.
- Render additionalProperties when it's an object.
- Render array items when it's an object.
- Change the location of the "Additional properties are supported" notice to make it clearer on nested objects.

### DIFF
--- a/library/src/containers/Schemas/SchemaProperties.tsx
+++ b/library/src/containers/Schemas/SchemaProperties.tsx
@@ -69,6 +69,29 @@ const renderProperties = (
   ));
 };
 
+const renderAdditionalProperties = (
+  schema: Schema,
+  treeSpace: number,
+): React.ReactNode => {
+  const additionalProperties = schema.additionalProperties;
+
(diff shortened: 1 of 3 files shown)


library/src/containers/Schemas/SchemaProperties.tsx
```diff
@@ -176,12 +203,124 @@ export const SchemaPropertiesComponent: React.FunctionComponent<Props> = ({
               must match {element.schema.content.pattern}
             </span>
           )}
+          {element.schema.content.uniqueItems && (
+            <span
```
fmvilas, line 207: I know these "spans" can be extracted to a component but don't want to lose time here refactoring since the React component itself will need a refactor and rethink soon.

derberg (approved):

magicmatatjahu:
@all-contributors please add @fmvilas for code, maintenance

PR #153, opened 2020-10-02
fix: let the list of enums break into multiple lines

Fox32:
**Description**

If we have a lot of enum values, the layout is broken because the enums are displayed in a single line. This can't be solved using CSS only, as there is no space between the enum values that the browser can break at. We add a spaced and remove the margin instead.

Before: 
image

After:

image


magicmatatjahu (approved):
Thanks for contribution! :)

derberg:
@all-contributors please add @Fox32 for code

PR #161, opened 2020-10-06
test: remove unnecessary undefined on test

### LINKED ISSUE #141: Remove not needed undefined from the toKebabCase test
**Context**

This react component is written with TS. SonarCloud states:

> Unlike in JavaScript, where every parameter can be omitted, in TypeScript you need to explicitly declare this in the function signature. Either you add ? in the parameter declaration and undefined will be automatically applied to this parameter. Or you add an initializer with a default value in the parameter declaration. In the latter case, when passing undefined for such parameter, default value will be applied as well. So it's better to avoid passing undefined value to an optional or default parameter because it creates more confusion than it brings clarity. Note, that this rule is only applied to the last arguments in function call.

Looking at the [implementation](https://github.com/asyncapi/asyncapi-react/blob/3130bf6b6a240734facc4601262ba3be033f2083/library/src/helpers/toKebabCase.ts#L1) you can see that we specify `str` argument with `?`. Passing `undefined` explicitly in the test is redundant and removing it should keep tests pass.

**Description**

Just remove `undefined` from [this](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/helpers/__tests__/toKebabCase.test.ts#L9) test.
Test description, instead of `should return empty string if input is undefined` should say something like `should return empty string if input is not provided`﻿

depimomo:
**Description**

Remove unnecessary undefined on test

**Related issue(s)**
Fixes #141

### DIFF
--- a/library/src/helpers/__tests__/toKebabCase.test.ts
+++ b/library/src/helpers/__tests__/toKebabCase.test.ts
@@ -6,7 +6,7 @@ describe('toKebabCase', () => {
   });
 
   test('should return empty string if input is undefined', () => {
-    expect('').toEqual(toKebabCase(undefined));
+    expect('').toEqual(toKebabCase());
   });
 
   test('should work with `lorem ipsum BIG-letter @!#$% something`', () => {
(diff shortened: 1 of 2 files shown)


magicmatatjahu (approved):
@depimomo Thanks for contribution! I changed title from `fix: ...` to `test: ...` because your PR hasn't any changes in code, but only in tests - no need to trigger the release.

magicmatatjahu:
@all-contributors please add @depimomo for test

PR #163, opened 2020-10-06
fix: Decrease package size by removing raml-dt parser from dependencies

### LINKED ISSUE #160: What is the recommended way to consume this component?
When consuming the component, I encountered a number of errors trying to build with webpack; and the resulting bundle size for my `vendor` bundle ended up at 4Mb (3.24Mb of this is `webapi-parser`).

Clearly I'm doing something wrong?

Here's how I did it:

1. Install the dependencies:
```
npm i @kyma-project/asyncapi-react react react-dom
```

2. Consume the component:
```ts
// app.tsx
import React from "react";
import { render } from "react-dom";
import AsyncApi, { ConfigInterface } from "@kyma-project/asyncapi-react";

const config: Partial<ConfigInterface> = {
  show: {
    schemas: false
  },
  showErrors: false
};

const schema = ` (omitted for brevity) `;

render(<AsyncApi schema={schema} config={config} />, document.getElementById("root"));
```

3. Build with webpack (`mode=production`)

The first build error I encountered was:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'fs' in 'node_modules/webapi-parser'
```

...which I was able to suppress by adding the following to my `webpack.config.js`:

```js
node: {
  fs: "empty"
},
```

The next error I encountered was:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'ajv/lib/refs/json-schema-v5.json' in 'node_modules/json-schema-migrate/lib'
```

It would seem that `json-schema-migrate` needs `ajv@5.x`; but one of the webpack plugins I use (`mini-css-extract-plugin`) needs `ajv@6.x` (which doesn't include the file `json-schema-v5.json`.

In my case, npm installed `ajv@5.x` in the `node_modules` folder inside `json-schema-migrate`, so I was able to suppress this error with a webpack alias:

```js
resolve: {
  alias: {
    "ajv/lib/refs/json-schema-v5.json": "json-schema-migrate/node_modules/ajv/lib/refs/json-schema-v5.json"
  },
```

Similarly, `markdown-it` requires the file `entities/maps/entities.json`, but the file path (on my machine) is actually `entities/lib/maps/entities.json`, so another webpack alias fixes that:

```
ModuleNotFoundError: Module not found: Error: Can't resolve 'entities/maps/entities.json' in 'node_modules/markdown-it/lib/common'
```

```js
resolve: {
  alias: {
    "ajv/lib/refs/json-schema-v5.json": "json-schema-migrate/node_modules/ajv/lib/refs/json-schema-v5.json",
    "entities/maps/entities.json": "entities/lib/maps/entities.json"
  },
```

With those errors fixed, I'm able to build successfully, but the size of the bundle is huge.

Here's what it looks like in `source-map-explorer`:

vendor-9dafdf_js_-_Source_Map_Explorer

Is this expected?

magicmatatjahu:
**Description**

- Decrease package size by removing `@asyncapi/raml-dt-schema-parser` from dependencies
- Remove `json-schema-ref-parser` package -> it's already defined in `@asyncapi/parser`.
- Remove unnecessary tests -> it makes not sense to test parser in component.

**Related issue(s)**
Resolves #160


derberg (changes requested):
you should only remove `@asyncapi/raml-dt-schema-parser` that is not browser compatible. Other parsers are needed as they are already used by the community

magicmatatjahu:
@derberg Done :)

library/src/config/config.ts
```diff
@@ -1,12 +1,11 @@
-import { Options as ParserOptions } from 'json-schema-ref-parser';
 import { PushStateBehavior } from '../types';
 
 export interface ConfigInterface {
   schemaID?: string;
   show?: ShowConfig;
   expand?: ExpandConfig;
   showErrors?: boolean;
-  parserOptions?: ParserOptions;
+  parserOptions?: any;
```
derberg, line 8: any ❤️

derberg (approved):
lgtm

derberg:
#160 will be solved but would be good to get a followup issue to work on a solution that custom parsers are optional and that users can register their own custom schema parsers

library/src/config/config.ts
magicmatatjahu: I checked that parser-js has different signature than ParserOptions from `json-schema-parser`

magicmatatjahu:
@derberg Adding custom parsers for asyncapi-react component will be easy if we decide to make parser per class (https://github.com/asyncapi/parser-js/issues/178)

Issue #168, opened 2020-10-13
Address XSS vulnerability in dompurify by updating dependency


renvrant:
This issue will be resolved by merging https://github.com/asyncapi/asyncapi-react/pull/135

magicmatatjahu:
@renvrant Hello :) I created a PR with grouped bump of deps with security issues here https://github.com/asyncapi/asyncapi-react/pull/169 Could you wait until tomorrow before the second codeowner (@derberg) does the review?

PR #169, opened 2020-10-13
fix: bump dependencies to resolve security issues

magicmatatjahu:
**Description**

- Bump deps in library, playground and in root to resolve security issues like https://github.com/asyncapi/asyncapi-react/issues/168
- Sleep in release workflow 10 seconds, before installing latest version in playground

**Related issue(s)**
Resolves https://github.com/asyncapi/asyncapi-react/issues/168

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,17 +37,13 @@ jobs:
           GIT_COMMITTER_NAME: asyncapi-bot
           GIT_COMMITTER_EMAIL: info@asyncapi.io
         run: npm run release
-      # Sleep for 10 seconds before using latest version in playground, because sometimes NPM needs additional few seconds to `save` package in registry
-      - name: Sleep for 10 seconds
-        uses: jakejarvis/wait-action@master
-        with:
-          time: '10s'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -37,6 +37,11 @@ jobs:
           GIT_COMMITTER_NAME: asyncapi-bot
           GIT_COMMITTER_EMAIL: info@asyncapi.io
         run: npm run release
+      # Sleep for 10 seconds before using latest version in playground, because sometimes NPM needs additional few seconds to `save` package in registry
+      - name: Sleep for 10 seconds
+        uses: jakejarvis/wait-action@master
```
derberg, line 42: please use something different, this action looks nifty but as you can see it is always fetched from master 😱 which is super unsecure. Unfortunately the user doesn't seem to provide regular professional releases

why not just `sleep 30s && npm install @kyma-project/asyncapi-react@${{ steps.extractver.outputs.version }} -s` in the playground step?

magicmatatjahu: Yes, you are right, I changed it to `sleep 10 && ...`

PR #172, opened 2020-10-22
fix: correct display of minimal value in case it is 0

derberg:
**Description**

- correct display of minimal value in case it is 0. The current condition could not work as 0 evaluates to undefined


fmvilas (approved):

library/src/containers/Schemas/SchemaProperties.tsx
```diff
@@ -250,7 +250,7 @@ export const SchemaPropertiesComponent: React.FunctionComponent<Props> = ({
               length &lt;= {element.schema.content.maxLength}
             </span>
           )}
-          {element.schema.content.minimum && (
+          {typeof element.schema.content.minimum === 'number' && (
```
magicmatatjahu, line 253: Could you also check `maximum` property? :)

magicmatatjahu (approved):

PR #185, opened 2020-11-16
fix: update version of web-component

magicmatatjahu:
**Description**

As in title: update version of web-component to `0.16.2`


derberg (approved):
one step for a man but a huge step for the tiger

PR #201, opened 2020-12-07
fix: refresh state in playground by debounce

### LINKED ISSUE #200: Fixed RefreshIcon location
**Description**
Set the refresh icon (on https://asyncapi.github.io/asyncapi-react/) to a fixed location so it's still visible when scrolling through YAML content on left hand pane.

**Reasons**
When debugging a large AsyncAPI document (few hundred lines+), manually scrolling to the top of the page each time to trigger an update is somewhat tedious. Having the refresh icon in a fixed location that's always visible regardless of scrolling would avoid this.

magicmatatjahu:
**Description**

- More info: https://github.com/asyncapi/asyncapi-react/issues/200
- Add `debounce` when user input new data in schema editor - after 750ms new view should be rendered with updated schema.
- Show refresh icon with spin animation when schema or config is updating.

> NOTE: `fix` is only added due to automatically push new playground.

New look:
image

**Related issue(s)**
Resolves #200

### DIFF
--- a/playground/src/Playground.tsx
+++ b/playground/src/Playground.tsx
@@ -33,9 +33,7 @@ class Playground extends Component<{}, State> {
   state = {
     schema: defaultSchema,
     config: defaultConfig,
-    schemaFromEditor: defaultSchema,
     schemaFromExternalResource: '',
-    configFromEditor: defaultConfig,
     refreshing: false,
   };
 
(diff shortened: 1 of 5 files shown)


derberg:
@magicmatatjahu refresh button is still there because of configuration? shouldn't it be visible only when you switch to configuration tab?

magicmatatjahu:
@derberg I left it because someone would like refresh manually state, but I find weird bug with memory allocation so I will remove icon as well.

magicmatatjahu:
@derberg I think everything should be done.

derberg (approved):
:shipit:

fmvilas (approved):

PR #213, opened 2020-12-18
fix: show 'const' values in property description (#210)

### LINKED ISSUE #210: show `const` values in property description
**Description**

If a property has a `const` value assigned, the const value is not shown as part of the rendered property description (it is shown as the value in `examples` though).  Swagger editor does for instance show `const` values.

**Expected result**

`const` value should be shown in the property description.

**Actual result**

`const` value is omitted from the property description.

**Steps to reproduce**

Add the following property to an arbitrary payload.

```json
"cityName": {
    "type": "string",
    "const": "Ratisbona"
}
```

**Troubleshooting**

I assume the `const` value would already be part of the parsed schema, so would adding the const to the SchemaPropertiesComponent would be sufficient?
https://github.com/asyncapi/asyncapi-react/blob/36d3a21dabedb0e8c74876b7626f9a8045901ea3/library/src/containers/Schemas/SchemaProperties.tsx#L174

Swagger editor:
image

AsyncApi react component
image

rexindan:
According to [JSON Schema specification](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.6.1.3)

> 6.1.3. const
> The value of this keyword MAY be of any type, including null.

So it's made `any` in 'types.ts'.
And because `const: ` can be `any` including having a value of `null`, I cannot check for existence of property `const: ` with `typeof el.schema.content.const === null` or `typeof el.schema.content.const === undefined` - 'cause these might be simply regular values of `const` (like `const: null` or `const: undefined`) saying nothing about its very existence. So I had to use `el.schema.content.hasOwnProperty('const') === true` to check that such property even exists in the object.

Every type transforms well to `String` (functions also) except `object`, so type `object` is serialized with `JSON.stringify` before rendering.

`{' '}` was added by Prettier, which adds it in React projects each time it doesn't like the length of the string, so can be removed forcefully if required.

Resolves #210

### DIFF
--- a/library/src/containers/Schemas/SchemaProperties.tsx
+++ b/library/src/containers/Schemas/SchemaProperties.tsx
@@ -151,6 +151,16 @@ const renderPropertyDescription = (el: SchemaElement): React.ReactNode => {
           Default: <span>{el.schema.content.default}</span>
         </div>
       )}
+      {el.schema.content.hasOwnProperty('const') && (
+        <div>
+          Const:{' '}
+          <span>
+            {typeof el.schema.content.const !== 'object'
+              ? String(el.schema.content.const)
+              : JSON.stringify(el.schema.content.const)}
+          </span>
+        </div>
+      )}
     </div>
   );
 };
--- a/library/src/types.ts
+++ b/library/src/types.ts
@@ -339,6 +339,7 @@ export interface Schema {
   required?: string[];
   enum?: any[];
   deprecated?: boolean;
+  const?: any;
   type?: SchemaType;
   items?: Schema /*| Schema[];*/; // todo: end this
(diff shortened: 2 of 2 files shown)


magicmatatjahu (commented):
@rexindan Thanks for contribution! Instead going with Swagger styling, we should go with styling in [html-template](https://github.com/asyncapi/html-template). You can check it in https://playground.asyncapi.io/ So please change `const:` to `Const:` and remove `<em>` tags like [here](https://github.com/asyncapi/html-template/blob/master/partials/schema-prop.html#L90).

rexindan:
Changed.

magicmatatjahu (approved):

PR #217, opened 2021-01-12
feat: showing required flag

c-pius:
**Description**

- showing a "required" flag for required properties (as also done in the [playground](https://playground.asyncapi.io/))

**Related issue(s)**

none


c-pius:
required-flag

the style is similar to how it is done in the [playground](https://playground.asyncapi.io/). Unfortunately I couldn't find out how to properly indent the "required" text if it is part of the tree. I assume it has something to do with the TreeSpace but couldn't really find out here https://github.com/asyncapi/asyncapi-react/blob/daf9f13440a8218713c8f678fae6b676c834c0cc/library/src/containers/Schemas/SchemaProperties.tsx#L125

c-pius:
the [playground](https://playground.asyncapi.io/) also shows a required flag for channel parameters. I assume this could be done as well by passing a required=true flag to the SchemaComponent and chaining it through:

https://github.com/asyncapi/asyncapi-react/blob/daf9f13440a8218713c8f678fae6b676c834c0cc/library/src/containers/Channels/Parameter.tsx#L26

https://github.com/asyncapi/asyncapi-react/blob/daf9f13440a8218713c8f678fae6b676c834c0cc/library/src/containers/Schemas/Schema.tsx#L38

should I do that as well?

magicmatatjahu:
@c-pius Hello! I don't have any problems with changed code. About indentations: exactly, you can use `TreeSpace` component like [here](https://github.com/asyncapi/asyncapi-react/blob/3d2f0284239f2e7773c28d0c8e2d0869d8c323bc/library/src/containers/Schemas/SchemaProperties.tsx#L131), so using `treeSpace` prop you add indentations in `for loop` :) About parameter: if you want, feel free to do it :)

c-pius:
@magicmatatjahu great. So parameters are now always shown as required. Indentation also works (it is not 100% aligned as you can see from the screenshot but I would consider it okay).

Screenshot 2021-01-14 at 10 26 34

magicmatatjahu:
@c-pius Thanks! Looks great :) If you want, you can manipulate with left padding to adjust the required to top text.

library/src/containers/Channels/Parameter.tsx
```diff
@@ -29,6 +29,7 @@ export const Parameter: FunctionComponent<Props> = ({
         schema={schema}
         hideTitle={true}
         description={description}
+        required={true} // parameters are always required
```
fmvilas, line 32: Good catch 👍

fmvilas (approved):
Thanks, @c-pius!

c-pius:
@magicmatatjahu
> If you want, you can manipulate with left padding to adjust the required to top text.

Couldn't really get it to work with padding-left so I inserted this one. Not sure if this is dirty? 

```
          if (el.treeSpace) {
            for (let i = 0; i < el.treeSpace; i++) {
              treeSpaces.push(<TreeSpace key={i} />);
            }
            // adding 25px span to account for the missing TreeLeaf
            treeSpaces.push(
              <span style={{ display: 'inline-block', width: '25px' }} />,
            );
          }
```

magicmatatjahu:
@c-pius Ok, you can leave it, yes it's dirty 😄 We don't want to make inline css, because component has a possibility to override css, and we shouldn't override this feature.

c-pius:
alright, reverted the dirty stuff 😄

c-pius:
> EDIT: First we want to merge this PR, before your one #218

sure, nor hurry.

derberg:
@c-pius please update your pr with latest master

@magicmatatjahu I merged, whenever you are ready with this one, approve and merge and let's see how this new CI beast works 😅

c-pius:
@derberg ready

magicmatatjahu (approved):

c-pius:
🙌🏻  thanks!

derberg:
@all-contributors please add @c-pius for code, bug

PR #218, opened 2021-01-13
ci: cleanup release workflow

derberg:
**This is a specific PR where we will not only introduce workflows that are core for this project, but also workflows that will be managed by global config in .github repo. The reason is = to test those workflows in a complex project before we know we can roll it out to others**

The files that you can ignore in review as they are marked with a comment as global and they are 1:1 taken from [this](https://github.com/asyncapi/.github/pull/21) and [that](https://github.com/asyncapi/.github/pull/18) PR:
- `.github/workflows/if-nodejs-pr-testing.yml`
- `.github/workflows/if-nodejs-release.yml`
- `.github/workflows/if-nodejs-version-bump.yml`
- `.github/workflows/issues-prs-notifications.yml`
- `.github/workflows/release-announcements.yml`
---
- issue of having specific `npm run bootstrap` solved with usage of npm lifecycle script, `postinstall` that is always triggered after installation of the package. So after doing `npm install` this is what I can see now
    ```
    wookiee@Lukaszs-MacBook-Pro asyncapi-react (master) $ npm install
    
    > undefined postinstall /Users/wookiee/sources/asyncapi-react
    > npm run bootstrap
    
    
    > @ bootstrap /Users/wookiee/sources/asyncapi-react
    > lerna bootstrap
    ```
    > don't worry about `undefined` , it is because root package.json has no version and no name

- issue with additional steps for web component and playground mitigated and moved to a separate workflow that reacts on release event
- issue with specific `npm run build-library` and `npm run build-webcomponent` mitigated with `prepublishOnly` script that triggers both
- and version bump can be done in global workflow for bumps by introducing 
    ```
        "bump:playground:version": "cd playground && npm --no-git-tag-version --allow-same-version version $VERSION",
        "bump:webcomp:version": "cd web-component && npm --no-git-tag-version --allow-same-version version $VERSION",
        "bump:lib:version": "cd library && npm --no-git-tag-version --allow-same-version version $VERSION",
        "bump:version": "npm run bump:lib:version && npm run bump:webcomp:version && npm run bump:playground:version"
    ```


derberg:
don't worry about failing job about PR notification - it is because on PR from fork I cannot modify PR workflow and expect to have access to secrets

.github/workflows/release-wc-and-playground.yml
```diff
@@ -0,0 +1,71 @@
...
+        uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - name: Install latest version of library in the web-component
+        # sleep for 60 seconds before using latest version in web-component, because sometimes NPM needs additional few seconds to `save` package in registry
+        run: |
+          sleep 1m
+          echo "Log all versions of the package on npm registry"
```
magicmatatjahu, line 24: Duplicated logs in playground and web-component are needed?

.github/workflows/release-wc-and-playground.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v2
+      - name: Setup Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: 14
+      - name: Install latest version of library in the web-component
+        # sleep for 60 seconds before using latest version in web-component, because sometimes NPM needs additional few seconds to `save` package in registry
+        run: |
+          sleep 1m
```
magicmatatjahu, line 23: As I understand, this script will run parallel with `if-nodejs-release.yml` and for that we have `1m` sleep? Another question: What about bumping versions? As I understand from scripts, we will have 2 extra commits with bumps on master: first for react-component, then for playground and web-component, yes?

.github/workflows/release-wc-and-playground.yml
derberg: there are 2 independent jobs that run at the same time, one can fail another not 🤷🏼‍♂️

magicmatatjahu: ok, thanks! I thought that those are a steps, not a jobs.

.github/workflows/release-wc-and-playground.yml
derberg: Sleep was there already, we needed it no matter if it is in the release workflow or here in the new one.
They do not run in parallel. The release is first as it is triggered by the push to master. This workflow starts when the last step of the release finish (GitHub release creation). In theory, no sleep needed, but you remember we had this problem with "moment of publish to npm" vs "when it is actually available for install"

.github/workflows/release-wc-and-playground.yml
derberg: @magicmatatjahu I want to use jobs as much as possible now for 2 reasons:
- config looks cleaner
- github has now this visualization feature that shows jobs and if they depend on each other. I mean, somebody spent time on implementing it, let us use it 😄

fmvilas (dismissed):

.github/workflows/release-wc-and-playground.yml
magicmatatjahu: So `if-nodejs-release.yml` and `release-wc-and-playground.yml` won't be triggered in this same time?

derberg: tl;dr no 😄 

`if-nodejs-release.yml` is triggered on `push` to master, and `release-wc-and-playground.yml` on `release`. So the 2nd one will be triggered once [this](https://github.com/asyncapi/asyncapi-react/pull/218/files#diff-8ed11ec3da81556149f1f4a379d4229113f23d9ae8c4eea626132bac8d535ea6R53-R63) creates a release


--- 
while writing this comment I realized `release-wc-and-playground.yml` is now triggered on `push` and not `release` 😅 pushing fix in a minute

derberg: Done 🥇

derberg:
@magicmatatjahu I just pushed a fix for the problem of not saving in git info about latest react component used in web component and playground. What do you think?

magicmatatjahu:
@derberg As I see, you run `install:reactcomp` in two workflows: `release-wc-and-playground` and here https://github.com/asyncapi/asyncapi-react/pull/218/files#diff-ef4f20ad9ac611c67466a8f25e69c3924ed476b297b5dd67f54829c012ea3a00R29 This is how it works?

derberg:
@magicmatatjahu yes, in the workflow about release it is needed for release purpose, but this information is not later pushed to the repo. This is what the other workflow does

derberg:
@magicmatatjahu if you are waiting for `Testing pull request` with approval, then don't as I removed this check with this PR 😄

magicmatatjahu (approved):

magicmatatjahu (approved):

PR #230, opened 2021-01-20
feat: add new props to web component and ease fonts handling

### LINKED ISSUE #191: Improve user experience with the web component
**Description**

- consider adding prop `schemaUrl`
- solve issue with fonts download without rethinking css in general. Maybe we can add fonts to `react-component` package to publish it to npm and have them hosted on unpkg, uncluding basic css that imports those fonts?

**Reasons**

to use web component as easily as described in [this](https://github.com/jfveronelli/instant-asyncapi-doc) project, I need to do this at the moment:

```
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
    <title>AsyncAPI Web Component Demo</title>
    <script src="https://unpkg.com/@asyncapi/web-component@0.17.5/lib/asyncapi-web-component.js" defer></script>
    <script>
    
        var schema = { url: "https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/simple.yml" };

        window.onload = function() {
            document.getElementById("asyncapi").schema=schema;
        } 
    </script>
  </head>
  <body>
    
    <asyncapi-component id="asyncapi" cssImportPath="https://unpkg.com/@asyncapi/react-component@0.17.5/lib/styles/fiori.css"></asyncapi-component>

  </body>
</html>
```

- custom script should not be necessary but this is how it works at the moment with the component, I need to pass object to the `schema`
- elements like toggle chevron for example or fonts are not properly rendered, I would have to add custom css like https://github.com/asyncapi/asyncapi-react/blob/master/playground/public/index.css and keep fonts in my project

aeworxet:
This PR covers several parts of the **asyncapi-web-component** (which is still a React component, though wrapped).

### Functionality

Current implementation allows `schemaUrl`/`schemaFetchOptions`/`schema.url`/`schema.requestOptions` to be specified in any combinations, whether as props of the Web Component
```html
<head>
</head>
<body>
  <asyncapi-component
    id="asyncapi"
    schemaUrl="https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/streetlights.yml"
    schemaFetchOptions='{"method":"GET","mode":"cors"}'
    cssImportPath="https://unpkg.com/@asyncapi/react-component@0.19.0/lib/styles/fiori.css">
  </asyncapi-component
</body>
```
or the way it was used before (as it was specified in the description of issue [https://github.com/asyncapi/asyncapi-react/issues/191](https://github.com/asyncapi/asyncapi-react/issues/191)), to ensure backward compatibility
```html
<head>
  <script>
    var schema = {
      url: "https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/simple.yml",
      requestOptions: {"method":"GET","mode":"cors"}
    };
    window.onload = function() {
      document.getElementById("asyncapi").schema=schema;
    };
  </script>
</head>
<body>
  <asyncapi-component
    id="asyncapi"
    cssImportPath="https://unpkg.com/@asyncapi/react-component@0.19.0/lib/styles/fiori.css">
  </asyncapi-component
</body>
```
or mixed (`schemaUrl`/`schema.requestOptions`, `schema.url`/`schemaFetchOptions`)
```html
<head>
  <script>
    var schema = {
      requestOptions: {"method":"GET","mode":"cors"}
    };
    window.onload = function() {
      document.getElementById("asyncapi").schema=schema;
    };
  </script>
</head>
<body>
  <asyncapi-component
    id="asyncapi"
    schemaUrl="https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/streetlights.yml"
    cssImportPath="https://unpkg.com/@asyncapi/react-component@0.19.0/lib/styles/fiori.css">
  </asyncapi-component
</body>
```
In all three cases the result will be the same.

If Web Component's props and Object's properties are specified simultaneously (different values are assigned both as a prop of the React component and as a property of object `schema` - and both at the same time)
```html
<head>
  <script>
    var schema = {
      url: "https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/simple.yml",
      requestOptions: { "method": "GET" }
    };
    window.onload = function() {
      document.getElementById("asyncapi").schema=schema;
    };
  </script>
</head>
<body>
  <asyncapi-component
    id="asyncapi"
    schemaUrl="https://raw.githubusercontent.com/asyncapi/asyncapi/master/examples/2.0.0/streetlights.yml"
    schemaFetchOptions='{"method":"GET","mode":"cors"}'
    cssImportPath="https://unpkg.com/@asyncapi/react-component@0.19.0/lib/styles/fiori.css">
  </asyncapi-component
</body>
```
the logic is as follows:

- if `schemaUrl` property is specified, the `schema.url` property of the inner React component will be ignored.
- if `schemaFetchOptions` property is specified, `schema.requestOptions` property of the inner React component will be ignored.
- if `schemaUrl` property is NOT specified, `schemaFetchOptions` will be ignored itself and `schema.url`/`schema.requestOptions` properties of the inner React component must be used in this case.

In case Web Component is called with no properties at all, error ("Error: The 'asyncapi' field is missing.") will be shown on page.

In case there are several Web Components on one page, each one will be rendered using its own properties.

#### Angular peculiarities

Angular 11.0.7 running in 'production mode + enforced stricter type checking and stricter bundle budgets' mode showed requirement of explicit definition of conditions, hence visually excessive code.

Both `undefined` and `'undefined'` are used because first has typeof `undefined` and the second has typeof `string`. And these differences do matter and work differently in current Angular's usecase, because Angular calls `AsyncApiWebComponent.prototype.render` multiple times for each property specified in function `register`, assigning primitive type `undefined` to undefined properties first time and assigning string `'undefined'` each next time.

### Styles
Only two icons from entire iconset of `SAP-icons` were used during last two years, and it was profitable to embed similar icons ([https://icon-icons.com/icon/chevron-right/116844](https://icon-icons.com/icon/chevron-right/116844), [https://icon-icons.com/icon/launch/117613](https://icon-icons.com/icon/launch/117613)) into CSS as SVG->base64, since no other icons are planned for addition in nearest future. Hence, 3 files `SAP-icons.*` are deleted, freeing up 493.5 KiB of download size.

`base64` format was chosen to avoid possible cross-browser incompatibilities due to different interpretation of raw SVG format by different browsers.

**Used tools:** [SVG minifyer](https://www.svgminify.com), [base64 encoder](https://base64.guru/converter/encode/file)

Typography of font `72` was not significantly different from the one of `sans-serif`, hence, use of `72` in CSS is changed to the use of `sans-serif`, and files `72-*` are also deleted, freeing up 140.0 KiB of download size.

### README.md
README.md was updated to reflect changes in functionality and styling.

Resolves #191


web-component/src/AsyncApiWebComponent.tsx
```diff
@@ -15,20 +17,107 @@ export class AsyncApiWebComponent extends React.Component<
...
+    // calls it only once) +1 more time at the end - assigning primitive type
+    // `undefined` to undefined properties first time and assigning string
+    // `'undefined'` each next time.
+    let schemaUrl;
+    let schemaFetchOptions;
+
     const finalCssImportPath =
       this.props.cssImportPath || 'assets/asyncapi.css';
 
+    const thisPropsSchema =
```
magicmatatjahu, line 37: I think that these options like `when schemaUrl is undefined but schema.url is not, then merge schema.requestOptions or use schemaFetchOptions prop` is unnecessary and creates a lot of if-logic. I think that if `schemaUrl` exists, then we should go only with `schemaUrl` and `schemaFetchOptions` props, and don't lookup for `schema` prop. So logic should looks like:

```js
const props = this.props;
let schemaUrl = props.schemaUrl;
let schemaFetchOptions = props.schemaFetchOptions;

if (schemaUrl === undefined || schemaUrl === "undefined") {
  schemaUrl = undefined;
}
if (schemaFetchOptions === undefined || schemaFetchOptions === "undefined") {
  schemaFetchOptions = undefined;
}

let schema = props.schema || {};
if (schemaUrl) {
   const schemaRequestOptions = schemaFetchOptions
      ? JSON.parse(JSON.stringify(schemaFetchOptions))
      : {};
   schema = { url: schemaUrl, requestOptions: schemaRequestOptions };
}
```

as you can see it is simpler and doesn't require a lot of if-statements that are difficult to understand :)

README.md
```diff
@@ -89,7 +89,35 @@ You can also run the Playground application locally by following [this](./docs/d
...
+- `schemaFetchOptions` property is an **optional** `object` of [RequestInit](https://microsoft.github.io/PowerBI-JavaScript/interfaces/_node_modules_typedoc_node_modules_typescript_lib_lib_dom_d_.requestinit.html) type in JSON format, specific to Web Component, containing additional fetching options. It is a wrapper for `schema.requestOptions` property in `schema` object, which are both needed only in case process of fetching from a URL is any different from a usual browser request,
+    > **NOTE**: If `schemaFetchOptions` property is specified, `schema.requestOptions` property of the React component will be ignored. If `schemaUrl` property is NOT specified, `schemaFetchOptions` will be ignored itself and `schema.url`/`schema.requestOptions` properties of the React component must be used in this case.
+- `cssImportPath` property is the path to styles. Default version from `unpkg.com` contains guaranteed minimum styling for the Web Component,
+- [webcomponentsjs](https://www.npmjs.com/package/@webcomponents/webcomponentsjs) is a series of polyfills to make code runnable in old browsers. It is **optional** if you do not intend to support any.
+
```
magicmatatjahu, line 105: If you apply in code my suggestion in another comment, then you should update also these part of Readme :)

magicmatatjahu (changes requested):
Awesome! tbh, I don't like these if-logic in web-component's `render` function. It's very hard to understand and I added suggestion how we can simplify this.

derberg:
@aeworxet next time please 🙏🏼 don't force push because the reviewer cannot identify what changes you made and needs to basically look at the entire PR again, instead of checking the fixes from the next commits

magicmatatjahu (approved):
LGTM :)

derberg:
@aeworxet Great stuff man, great stuff! `schemaUrl` is a key element here to make it crazy easy to work with the component

derberg:
@all-contributors please add @aeworxet for doc, code

PR #256, opened 2021-03-02
feat: render bindings in relevant sections

195858:
**Description**

- Created a new bindings/binding component
-  Added styles for the new component
- Updated the types to include bindings in the appropriate interfaces
- Updated server, channel, operation and message components
to render bindings if present
- Slightly changed operation/messages rendering

### DIFF
--- a/library/src/constants.ts
+++ b/library/src/constants.ts
@@ -55,6 +55,11 @@ export const MESSAGE_PAYLOAD_TEXT = 'Message Payload';
 export const PAYLOAD_EXAMPLE_TEXT = 'Example of payload';
 export const SCHEMA_EXAMPLE_TEXT = 'Example';
 
+export const SERVER_BINDINGS_TEXT = 'Server Bindings';
+export const CHANNEL_BINDINGS_TEXT = 'Channel Bindings';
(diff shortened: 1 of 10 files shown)


magicmatatjahu:
@195858 Hi! Thanks for contribution! I will do a review by Friday at the latest :)

195858:
Thanks, when reviewing the changes you will realise I adjusted the way operations and its messages are rendered in a channel. The reason for that was that the "RawMessage" case for an operation (as opposed of having multiple messages associated) wasn't rendered that clearly and it was hard to figure out what's what in case there's a publish and subscribe operation with a single (RawMessage) option only.

derberg:
@195858 please update this PR by extending the streetlight example with bindings on all possible levels (the spec that is loaded by default in the playground). We always try to extend those example to show new features in action out-of-the-box

195858:
@derberg - streetlights spec update should now show in the pull request.

195858:
I just realised there are cases my changes don't handle - nested properties and SchemaObject propeerties.
I'll submit another PR once it's fixed.

magicmatatjahu:
@195858 Don't worry, If you have any problems, please ping us on the Slack :)

PR #264, opened 2021-03-17
refactor: switch to parser-js in most places 

magicmatatjahu:
**Description**

> if someone think that the PR is too big, I will try to divide the PR into two smaller ones.

- Switch to `parser-js` in `Info`, `Servers` (also in single `Server`), `Channels` (also in single `Channel`), `Messages` (also in single `Message`), `Schemas` components,
- left single `Schema` component, because it is the most complicated component in package and it should be rewritten with https://github.com/asyncapi/shape-up-process/issues/87 task - also by this I left some code related to "beautify" our schema - if `Schema` component will be rewritten, we will remove about leftover 10-15% of code.
- I wanna treat PR as chore, because I wanna merge to `unify-component` branch :)

**Related issue(s)**
Part of https://github.com/asyncapi/shape-up-process/issues/85

### DIFF
--- a/library/src/containers/AsyncApi/AsyncApi.tsx
+++ b/library/src/containers/AsyncApi/AsyncApi.tsx
@@ -123,9 +123,8 @@ class AsyncApiComponent extends Component<AsyncApiProps, AsyncAPIState> {
                 <ErrorComponent error={error} />
               )}
               {concatenatedConfig.show.info && <InfoComponent />}
-              {concatenatedConfig.show.channels && validatedSchema.channels && (
+              {concatenatedConfig.show.channels && (
                 <ChannelsComponent
-                  channels={validatedSchema.channels}
                   expand={
                     concatenatedConfig.expand &&
                     concatenatedConfig.expand.channels
--- a/library/src/containers/AsyncApi/AsyncApi.tsx
+++ b/library/src/containers/AsyncApi/AsyncApi.tsx
@@ -1,4 +1,5 @@
 import React, { Component } from 'react';
+import { AsyncAPIDocument } from '@asyncapi/parser';
 
 import {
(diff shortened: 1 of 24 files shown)


fmvilas (approved):
Great stuff, Maciej! 👏

fmvilas:
BTW, I think the [beautifying process](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/helpers/beautifier.ts) should be removed entirely. Not on this PR but at some point during the cycle. As it is right now, there are a couple of things that are not needed or incorrect:

* Resolving `allOf`: this is incorrect, we're treating `allOf` as merging objects and that's not the correct behavior.
* Rendering Markdown can be done using a React Markdown component.

magicmatatjahu:
@fmvilas

> BTW, I think the beautifying process should be removed entirely. Not on this PR but at some point during the cycle. As it is right now, there are a couple of things that are not needed or incorrect:
>
> Resolving allOf: this is incorrect, we're treating allOf as merging objects and that's not the correct behavior.
Rendering Markdown can be done using a React Markdown component.

I mention about removing this process in sentence:

> ...also by this I left some code related to "beautify" our schema - if Schema component will be rewritten, we will remove about leftover 10-15% of code.

I used React Markdown component in previous job and it's awesome, but this is a nice-to-have feature in unification :) We can stay with `markdown-it` library used here https://github.com/asyncapi/asyncapi-react/blob/master/library/src/helpers/renderMarkdown.tsx

derberg:
@magicmatatjahu I would just change commit scope to `refactor:` and :shipit:

magicmatatjahu:
@derberg Done :)

derberg (approved):

PR #269, opened 2021-03-22
refactor: rewrite Schema component

magicmatatjahu:
**Description**

Rewrite Schema component:
- Add new `NewSchema` component which will replace in next PRs old `Schema` component - I don't want to rewrite everything in one PR
  - support for combined schemas like `oneOf`, `anyOf`, `allOf`
  - handle `additionalProperties` and `additionalItems`
  - handle circular properties
- Start rewriting styles to TailwindCSS - at the moment only used in `NewSchema` component.
- Add `dummy` spec to Playground.

New look:

image

I know that some styles don't look nice, but this PR is only to "switch" to new way of rendering, styling will be done as part https://github.com/asyncapi/shape-up-process/issues/86 in next PR.

**Related issue(s)**
Part of https://github.com/asyncapi/shape-up-process/issues/87
Part of https://github.com/asyncapi/shape-up-process/issues/86


library/src/containers/Schemas/NewSchema.tsx
```diff
@@ -0,0 +1,345 @@
...
+
+import { Chevron, Markdown } from '../../components';
+import { SchemaHelpers } from '../../helpers';
+
+interface Props {
+  schemaName?: string;
+  schema?: Schema;
+  required?: boolean;
+  isCircular?: boolean;
+  odd?: boolean;
```
fmvilas, line 12: `odd` looks weird to me. What does it mean that a prop is odd? (I know what it means, just want you to think about it from a different perspective).

Should this be called something like `darkBackground` or `lightBackground` or something like this? Something that refers more to what it actually does.

fmvilas (commented):
Looking great in general. I'll help you with the styles 😄

library/src/containers/Schemas/NewSchema.tsx
fmvilas: Maybe this light/dark background thingy can be handled outside this component? For instance, the NewSchema component is transparent and the color of the background is handled in the parent component.

magicmatatjahu: The `odd` I "copied" from HTML template and you probably understood what it is - it "checks" that another schema renders in another nested schema like `oneOf` or `object` as prop in another `object`. Of course we can call this `nested` or something like that, but HTML template do it in proper way - odd indicated also how nested schema is. `darkBackground` or `lightBackground` is good, but we should choose only one. 

> Maybe this light/dark background thingy can be handled outside this component? For instance, the NewSchema component is transparent and the color of the background is handled in the parent component.

The main problem with it are very nested schemas and style "wrappers" for them. Transparent of schema should works only for 'root' schema, not for nested. I will try to check something like:

```css
.schema > .schema {
 // light styles
}

.schema > .schema > .styles {
 // dark styles
}
```

magicmatatjahu: Ok, I removed `odd` prop and we make "odd" styles by pure css combinators.

magicmatatjahu:
@fmvilas I think that you haven't problem with logic, but you had one, which i corrected. Let me merge, because PR itself blocks the rest of the things - channels and messages parts :) As if there are any bugs, we will fix them in next PRs.

fmvilas:
No, I have no other problems than the `odd` stuff. I'll jump on fixing what you did "the Tailwind way" instead of this hacky CSS 😄

PR #270, opened 2021-03-22
feat: ability to change the size of the split panes

### LINKED ISSUE #268: Add the ability to change the size of the split panes
**Description**

**This is related to the playground**. Add the ability to resize the split panes(the vertical line in the middle).

**Reasons**

Right now the editor window is kind of small, so the ability to resize it will give the power to users to resize it as they desire.

**Attachments**

To make it more clear, this vertical line(separation) is what I'm talking about:
vertical line

If the maintainers approve this feature request, I might be able to work on it.

aayushmau5:
**Description**

- Added the ability to change the size of split panes
- Used a library called `react-split`

**Related issue(s)**

Fixes #268

### DIFF
--- a/playground/src/components/SplitWrapper.tsx
+++ b/playground/src/components/SplitWrapper.tsx
@@ -1,0 +1,38 @@
+import React, { Component } from 'react';
+import Split from 'react-split';
+
+export interface WrapperProps {
+  children: React.ReactNode;
+}
+
+class SplitWrapper extends Component<WrapperProps> {
+  render() {
+    return (
+      <Split
+        style={{
+          width: '100%',
+          height: 'calc(100vh - 50px)',
+          minHeight: 'calc(100vh - 50px)',
+          display: 'flex',
+          background: '#f3f4f5',
+        }}
+        gutter={() => {
+          const gutter = document.createElement('div');
+          gutter.onmouseover = () => (gutter.style.cursor = 'ew-resize');
+          return gutter;
+        }}
+        gutterStyle={() => {
+          return {
+            backgroundColor: 'gray',
+            width: '7px',
+          };
+        }}
+        minSize={250}
+      >
+        {this.props.children}
+      </Split>
+    );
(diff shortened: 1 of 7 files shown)


playground/src/components/SplitWrapper.tsx
```diff
@@ -0,0 +1,36 @@
...
+        })}
+        minSize={250}
+      >
+        {this.props.children}
+      </Split>
+    );
+  }
+}
+
+export default SplitWrapper;
```
magicmatatjahu, line 36: You can use here normal FunctionComponent :)

```suggestion
import React from 'react';
import Split from 'react-split';

const SplitWrapper: React.FunctionComponent = ({ children }) => {
    return (
      <Split
        style={{
          width: '100%',
          height: 'calc(100vh - 50px)',
          minHeight: 'calc(100vh - 50px)',
          display: 'flex',
          background: '#f3f4f5',
        }}
        gutter={() => {
          const gutter = document.createElement('div');
          gutter.onmouseover = () => (gutter.style.cursor = 'ew-resize');
          return gutter;
        }}
        gutterStyle={() => ({
          backgroundColor: 'gray',
          width: '7px',
        })}
        minSize={250}
      >
        {this.props.children}
      </Split>
    );
}

export default SplitWrapper;
```

magicmatatjahu (commented):
Awesome! Please refer to my one suggestion :)

aayushmau5:
@magicmatatjahu Done :D

magicmatatjahu (approved):

magicmatatjahu:
@aayushmau5 Thanks for contribution!

derberg:
@magicmatatjahu I think we need to consider using scopes in future commit messages to indicate that change is for example only for the Playground, or only for web component. Although we still need it in commit text so it is clear in the release notes

derberg:
@all-contributors please add @aayushmau5 for code

PR #276, opened 2021-03-24
refactor: unify Messages and Operations components

magicmatatjahu:
**Description**

Unify `Messages` and `Operations` components,
- styles as in previous PR with new `Schema` component is not related to this PR but to https://github.com/asyncapi/shape-up-process/issues/86 task,
- render `correlationId` and `tags` in messages and operations.

> **NOTE**: there are still `Servers` and `Info` to rewrite, as well as custom extensions and bindings. Then we can start styling :)

Examples:

image

image

**Related issue(s)**
Part of https://github.com/asyncapi/shape-up-process/issues/87

### DIFF
--- a/library/src/components/Tags.tsx
+++ b/library/src/components/Tags.tsx
@@ -1,8 +1,8 @@
 import React from 'react';
-import { Tag } from '@asyncapi/parser';
+import { Tag as TagType } from '@asyncapi/parser';
 
 interface Props {
-  tags?: Tag[];
+  tags?: TagType[];
 }
 
 export const Tags: React.FunctionComponent<Props> = ({ tags }) => {
@@ -13,17 +13,13 @@ export const Tags: React.FunctionComponent<Props> = ({ tags }) => {
   return (
     <div className="mt-4">
       {tags.map(tag => (
-        <TagComponent tag={tag} />
+        <Tag tag={tag} />
       ))}
     </div>
   );
 };
 
-interface TagProps {
-  tag: Tag;
-}
-
-const TagComponent: React.FunctionComponent<TagProps> = ({ tag }) => {
+const Tag: React.FunctionComponent<{ tag: TagType }> = ({ tag }) => {
   const externalDocs = tag.externalDocs();
 
   if (externalDocs) {
--- a/library/src/components/Tags.tsx
+++ b/library/src/components/Tags.tsx
@@ -13,7 +13,7 @@ export const Tags: React.FunctionComponent<Props> = ({ tags }) => {
   return (
     <div className="mt-4">
       {tags.map(tag => (
-        <Tag tag={tag} />
+        <Tag tag={tag} key={tag.name()} />
       ))}
     </div>
   );
--- a/library/src/components/Tags.tsx
+++ b/library/src/components/Tags.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import { Tag as TagType } from '@asyncapi/parser';
 
+import { Tag } from './Tag';
+
 interface Props {
   tags?: TagType[];
 }
@@ -16,30 +18,5 @@ export const Tags: React.FunctionComponent<Props> = ({ tags }) => {
         <Tag tag={tag} key={tag.name()} />
       ))}
     </div>
-  );
-};
-
-const Tag: React.FunctionComponent<{ tag: TagType }> = ({ tag }) => {
-  const externalDocs = tag.externalDocs();
-
-  if (externalDocs) {
-    return (
-      <a
-        href={externalDocs.url()}
-        target="_blank"
-        className="border text-blue-400 font-normal text-sm rounded lowercase mr-2 px-2 py-1"
-        title={tag.description() || ''}
-      >
-        {tag.name()}
-      </a>
-    );
-  }
-  return (
-    <span
-      className="border font-normal text-sm no-underline text-blue-400 rounded lowercase mr-2 px-2 py-1"
-      title={tag.description() || ''}
-    >
-      {tag.name()}
-    </span>
   );
 };
--- a/library/src/components/Tags.tsx
+++ b/library/src/components/Tags.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import { Tag } from '@asyncapi/parser';
+
+interface Props {
+  tags?: Tag[];
+}
+
+export const Tags: React.FunctionComponent<Props> = ({ tags }) => {
+  if (!tags || !tags.length) {
+    return null;
+  }
+
+  return (
+    <div className="mt-4">
+      {tags.map(tag => (
+        <TagComponent tag={tag} />
+      ))}
+    </div>
+  );
+};
+
+interface TagProps {
+  tag: Tag;
+}
+
+const TagComponent: React.FunctionComponent<TagProps> = ({ tag }) => {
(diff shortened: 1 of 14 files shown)


library/src/components/Tags.tsx
```diff
@@ -0,0 +1,45 @@
...
+  return (
+    <div className="mt-4">
+      {tags.map(tag => (
+        <Tag tag={tag} key={tag.name()} />
+      ))}
+    </div>
+  );
+};
+
+const Tag: React.FunctionComponent<{ tag: TagType }> = ({ tag }) => {
```
fmvilas, line 22: I usually expect a component per file. Also, this component is not exported so I'd not be able to render it alone in Storybook. I think you should move this component to another file.

library/src/containers/AsyncApi/AsyncApi.tsx
```diff
@@ -182,18 +168,18 @@ class AsyncApiComponent extends Component<AsyncApiProps, AsyncAPIState> {
...
-      return null;
-    }
-    return beautifier.beautify(schema);
-  }
+  // private beautifySchema(schema: NullableAsyncApi): NullableAsyncApi {
+  //   if (!schema) {
+  //     return null;
+  //   }
+  //   return beautifier.beautify(schema);
+  // }
```
fmvilas, line 182: You probably can remove this.

library/src/containers/Channels/NewOperation.tsx
```diff
@@ -0,0 +1,78 @@
...
+          schema={parameters}
+          notRender={{
+            rootType: false,
+            additionalInfo: false,
+          }}
+        />
+      )}
+
+      <Markdown>{channel.description()}</Markdown>
+      <Markdown>{operation.summary()}</Markdown>
```
fmvilas, line 56: [Operation.summary](https://github.com/asyncapi/spec/blob/master/spec/asyncapi.md#operationObject) does not support Markdown.

library/src/containers/Channels/NewOperation.tsx
```diff
@@ -0,0 +1,78 @@
...
+      <Markdown>{operation.description()}</Markdown>
+
+      {operation.hasMultipleMessages() ? (
+        <div>
+          <p>
+            Accepts <strong>one of</strong> the following messages:
+          </p>
+          {operation.messages().map((msg, idx) => (
+            <MessageComponent message={msg} index={idx} key={idx} />
```
fmvilas, line 65: Why `MessageComponent` and not `Message`?

library/src/containers/Channels/NewOperations.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+  const className = CONTAINER_LABELS.CHANNELS;
+  const header = <h2>{CHANNELS_TEXT}</h2>;
+
+  const operationsList: React.ReactNodeArray = [];
+  Object.entries(channels).forEach(([channelName, channel]) => {
+    if (channel.hasPublish()) {
+      operationsList.push(
+        <li key={channelName}>
+          <OperationComponent
```
fmvilas, line 26: Why `OperationComponent` and not `Operation`?

library/src/containers/Channels/NewOperations.tsx
```diff
@@ -0,0 +1,68 @@
...
+            operationType={PayloadType.PUBLISH}
```
fmvilas, line 27: It's an operation component so I think this could just be named as `type`.

library/src/containers/Channels/NewOperations.tsx
```diff
@@ -0,0 +1,68 @@
...
+          />
+        </li>,
+      );
+    }
+  });
+
+  return (
+    <section
+      className={bemClasses.element(className)}
+      id={bemClasses.identifier([className])}
```
fmvilas, line 52: Why are you still using BEM in the new component? It should be using Tailwind instead.

library/src/containers/Messages/NewMessages.tsx
```diff
@@ -0,0 +1,47 @@
...
+          <MessageComponent message={msg} index={idx + 1} key={idx} />
+        </li>
+      ))}
+    </ul>
+  );
+
+  return (
+    <section
+      className={bemClasses.element(className)}
+      id={bemClasses.identifier([className])}
```
fmvilas, line 33: Same thing with BEM here.

fmvilas (changes requested):

library/src/components/Tags.tsx
magicmatatjahu: I decided that rendering a single tag is pointless and treated it as a sub-component of the Tags component, but I will move it to separate file for you 😘

library/src/containers/AsyncApi/AsyncApi.tsx
magicmatatjahu: I commented it on the first com,it and forgot about it 😅  thanks!

library/src/containers/Channels/NewOperation.tsx
magicmatatjahu: I know this, but I am wondering whether to render it as a markdown -> the content will be automatically wrapped in a paragraph, and then we can reuse markdown styling for summary. However, I will move it to normal `<p>`

library/src/containers/Channels/NewOperations.tsx
magicmatatjahu: Old nomenclature in all components (and in Kyma) :) But you're right, we need to switch to names without `Component` in the name.

library/src/containers/Channels/NewOperation.tsx
magicmatatjahu: Old nomenclature in all components (and in Kyma) :) But you're right, we need to switch to names without `Component` in the name.

library/src/components/Tags.tsx
fmvilas: hahaha! it's not pointless. It is pointless in your context but may not be when we present the design sytem.

library/src/containers/Channels/NewOperation.tsx
fmvilas: Yes, if something doesn't support Markdown per spec, we shouldn't give the option to people. Otherwise, their spec will be rendered differently depending on the tool.

magicmatatjahu: You're right :)

library/src/containers/Channels/NewOperations.tsx
magicmatatjahu: I know, but section only wraps content like this:

image

I wanna make styles here asyncapi/shape-up-process#86

library/src/containers/Messages/NewMessages.tsx
magicmatatjahu: I know, but section only wraps content like this:

image

I wanna make styles here asyncapi/shape-up-process#86

library/src/containers/Channels/NewOperations.tsx
fmvilas: Oh ok! I'll take care of this task if you don't mind.

magicmatatjahu:
@fmvilas Do you have another suggestions? I applied changes :)

fmvilas (approved):
:shipit:

PR #277, opened 2021-03-25
refactor: rewritten Info component

aeworxet:
This PR brings React component's playground appearance closer to [playground.asyncapi.io](https://playground.asyncapi.io)'s.
Unification process is in progress, CSS styles will be polished further.

### DIFF
--- a/library/src/components/Markdown.tsx
+++ b/library/src/components/Markdown.tsx
@@ -2,8 +2,6 @@ import React from 'react';
 import DOMPurify from 'dompurify';
 import MarkdownIt from 'markdown-it';
 
-import { bemClasses } from '../helpers';
-
 const markdownIt = new MarkdownIt();
 
 export const Markdown: React.FunctionComponent = ({ children }) => {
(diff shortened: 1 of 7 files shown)


library/src/components/Markdown.tsx
```diff
@@ -2,19 +2,17 @@ import React from 'react';
...
 import MarkdownIt from 'markdown-it';
 
-import { bemClasses } from '../helpers';
-
 const markdownIt = new MarkdownIt();
 
-export const Markdown: React.FunctionComponent = ({ children }) => {
+interface Props {
+  children?: string | null;
+}
```
magicmatatjahu, line 9: You don't need to define `children` prop for component :) Please remove it, or maybe write why you want to do that?

library/src/containers/Info/NewInfo.tsx
```diff
@@ -0,0 +1,54 @@
...
+import { useSpec } from '../../store';
+
+interface Props {
+  asyncapi: string;
+  info?: string;
+  termsOfService?: string;
+  contact?: string;
+  license?: string;
+  defaultContentType?: string;
+}
```
magicmatatjahu, line 17: You can remove this Props :)

magicmatatjahu (changes requested):
Awesome! Only two suggestions! :)

magicmatatjahu (approved):
Awesome! Great for contribution :)

Issue #265, opened 2021-03-17
Unify React component with HTML template - discussion


magicmatatjahu:
We're currently maintaining two HTML documentation views: [the HTML template](https://github.com/asyncapi/html-template/) and this React component. This is a big problem, because we must maintain two source code, which are do this same - render AsyncAPI spec as HTML. Of course users can use React component in SPA application and for dynamic data, and HTML template renders static content, but purpose is this same.

Some advantages with unification:
1. Less code to maintain.
2. We have 3 playground - [playground for React component](https://asyncapi.github.io/asyncapi-react/), [playground for generator](https://playground.asyncapi.io/) and new [studio](https://github.com/asyncapi/studio) which is in early dev stages. After this task, we will replace these 3 playground only with one - Studio. (go to 1. point) 
3. Unificated styles across official ecosystem of AsyncAPI - in website, Studio and React component. React component uses by default Fiori styling, but studio and website use [TailwindCSS](https://tailwindcss.com/).

I think that 1. and 2. points for everyone is very good and no one should have a problem with it (if you have, please write about your point of view). The most problematic is point 3. At the moment we have "theming" feature inside React component. User can override styling by overriding styles for appropriate css classes in this [file](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/styles/fiori.css). We know that some companies/users replaces default styling with new, related to internal design system of company etc, but we wanna know what impact will it have on your products that we will change the default styling and remove currently "theming" system for the next 2 months? Our priority is to unify styles and code across the official tooling ecosystem, and then add features to the component, such as theming or plugins. Please share your thoughts!

For active HTML template's contributors:
- What do you think about this movement? Do you feel good that for contribution to next versions of HTML template, you will write in React? Of course your contributions won't be removed 😄 but the "way" to add new contributions. To be honest... your code will be removed, I mean whole Nunjucks templates and for someone it can be a problem. Please share your thoughts!

For more info of each step of unification, please check tasks:
- [Unify HTML template and React component](https://github.com/asyncapi/shape-up-process/issues/83#positive-side-effects) - reference to shape up repo,
- [Use parser-js in react-component](https://github.com/asyncapi/shape-up-process/issues/85) - we wanna ride of inline retrieving of spec data like `spec.info.contact.email` etc and go to [parser-js](https://github.com/asyncapi/parser-js) api,
[...]

PR #296, opened 2021-04-08
refactor: unify styling

magicmatatjahu:
**Description**

Unify styling in component with HTML-Templates styling:
- styling `Info`, `Server`, `Operation`, `Message`, `Schema` components
- the number of lines is related with new dependencies like `postcss`, `tailwind` etc. The rest of the code is converting inline styles/classes into mixed classes.

> In next PR will be styling for Examples, Sidebar and for mobiles.

Examples:

- Info:
image

- Server 
image

- Operation 
image

- Message 
image

**Related issue(s)**
Part of https://github.com/asyncapi/shape-up-process/issues/86


library/src/styles/styles.css
```diff
@@ -0,0 +1,371 @@
...
+  .ai-info__title {
+    @apply text-3xl;
+  }
+  .ai-info__links {
+    @apply leading-normal mt-2 space-x-2 space-y-2;
+  }
+  .ai-info__links-item {
+    @apply inline-block;
+  }
+  .ai-info__links-item > a {
```
fmvilas, line 26: If we're going to choose BEM over utility-first directly on the component, we should strictly avoid using CSS operators like `>`. They will become a problem if someone wants to override the styles applied here as our CSS selector will always have more specificity than theirs.

And honestly, I'd use `@apply` only for those highly repeated styles (like buttons, links, etc.). I'd keep the styles in the component files otherwise.

fmvilas (changes requested):

library/src/styles/styles.css
magicmatatjahu: > If we're going to choose BEM over utility-first directly on the component, we should strictly avoid using CSS operators like >. They will become a problem if someone wants to override the styles applied here as our CSS selector will always have more specificity than theirs.

I agree with it partially 😄 People will also be able to override styles by `ai-info__links-item > a`. Funny fact: Tailwind also uses operators in some classes, especially in `display` etc.

> And honestly, I'd use @apply only for those highly repeated styles (like buttons, links, etc.). I'd keep the styles in the component files otherwise.

And then people won't be able to override styles for particular parts :)

fmvilas: > People will also be able to override styles by `ai-info__links-item > a`

Have you ever tried overriding Bootstrap styles? What do you think is the purpose of BEM? 😄 By doing this, you're breaking BEM and its benefits.

> And then people won't be able to override styles for particular parts :) I based styles on `@apply` directive having this problem in mind.

We don't want people to fully customize the components and all the CSS (unless they override CSS which is always possible). If someone wants to create a fully customized component, they'll be able to replace our components: https://github.com/fmvilas/asyncapi-pitches/blob/master/ui-library.md#customization. Not a problem to solve now.

In my experience, we should go the utility-first approach and maybe use some names for the colors, like primary, secondary, link, text, etc. Then people will be able to customize these colors. The same applies to sizes, paddings, margins, etc. It is, we must create a design system. We don't create CSS classes for the components but for the system the components are adhereing to.

magicmatatjahu: > We don't want people to fully customize the components and all the CSS (unless they override CSS which is always possible). If someone wants to create a fully customized component, they'll be able to replace our components: https://github.com/fmvilas/asyncapi-pitches/blob/master/ui-library.md#customization. Not a problem to solve now.

I go to link and read: `It should be easy to customize the components. Not just at CSS level but also at markup level.` 😄 I'm just kidding. 

@fmvilas I like the idea of ​​utility classes for buttons, borders, etc., and even today in the morning I noticed the problem that there are too many classes. The question is whether to start writing these classes, already or wait for the next cycle?

fmvilas: I'd not move these classes to the CSS file in this cycle. It's a strong decision that will affect customization. In future cycles, we'll work on customization and I think it's then when we have to make a choice.

magicmatatjahu: I will remember! 😄 Style customization is most important thing after component unification. We talked about it under the main task of the cycle, but I found that not only Kyma has his own styles, but also people from Spotify or one person from https://github.com/asyncapi/asyncapi-react/issues/265 issue. The old component has css theming feature and we break it now.

fmvilas: I don't like the styles. Not because of your changes but because the current design of the HTML template is broken and feels patched 😄 So yeah, I like the changes you did 😄

magicmatatjahu:
@fmvilas I moved the styles to components files :) Could you check again? I also moved those that are repeated, such as for buttons, because you are right that it is better to think about theming in the next cycles.

fmvilas (approved):
Looking great 👍 Thanks for taking my opinion into consideration! 🙏

PR #318, opened 2021-04-26
feat: bump dependencies before pushing new version of component

magicmatatjahu:
**Description**

As in title. We wanna bump all dependencies before pushing new version of component. Also fix rendering of bindings in web-component.

> `feat:` not `fix:` because we wanna start from `0.24.0` (or from `1.0.0-alpha`) new version of component.

**Related issue(s)**
Pat of https://github.com/asyncapi/shape-up-process/issues/83

### DIFF
--- a/library/package.json
+++ b/library/package.json
@@ -42,8 +42,7 @@
     "test": "jest --detectOpenHandles",
     "testWatch": "jest --detectOpenHandles --watch",
     "copyCss": "cp -r ./src/styles ./lib",
-    "prepare": "npm run build",
-    "prepack": "npm run build && cp ../README.md ./README.md && cp ../LICENSE ./LICENSE",
+    "prepare": "npm run build && cp ../README.md ./README.md && cp ../LICENSE ./LICENSE",
     "postpack": "rm -rf ./README.md && rm -rf ./LICENSE"
   },
   "dependencies": {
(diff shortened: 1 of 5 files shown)


library/package.json
```diff
@@ -43,7 +43,7 @@
     "testWatch": "jest --detectOpenHandles --watch",
     "copyCss": "cp -r ./src/styles ./lib",
     "prepare": "npm run build",
-    "prepack": "cp ../README.md ./README.md && cp ../LICENSE ./LICENSE",
+    "prepack": "npm run build && cp ../README.md ./README.md && cp ../LICENSE ./LICENSE",
```
magicmatatjahu, line 46: We don't push Readme to npm from some commits. This line fixes it.

package.json
```diff
@@ -17,7 +17,7 @@
     "start": "lerna exec --parallel -- npm run start",
     "test": "cd library && npm test",
     "build": "npm run build:library && npm run build:webcomponent && npm run build:playground",
-    "build:library": "cd library && npm run prepare",
+    "build:library": "cd library && npm run prepack",
```
derberg, line 20: what is wrong with `prepare`? why this has to change?

magicmatatjahu: `prepare` is run with `npm run bootstrap` which is `lerna bootstrap`. When you publish library from root `package.json` you execute `prepare` without `prepack` and we don't have Readme.md in npm https://www.npmjs.com/package/@asyncapi/react-component?activeTab=dependencies

derberg: isn't it better to extend `prepare` than now maintain both, `prepare` and `prepack`, don't you think it will be confusing and lead to more problems?

magicmatatjahu: ok then.

derberg (approved):

PR #320, opened 2021-04-26
feat: unify component with html-template

magicmatatjahu:
**Description**

Unify component with with html-template. For more info see https://github.com/asyncapi/shape-up-process/issues/83 and https://github.com/asyncapi/asyncapi-react/issues/265

Changes:
- remove Fiori styling and start using TailwindCSS - theming at the moment won't be possible, we want to add it in the future in simplifier form that currently - see https://github.com/asyncapi/shape-up-process/issues/83 -> Don't support theming section. `@asyncapi/react-component/lib/styles/fiori.css` is deprecated.
- default styles now is in `@asyncapi/react-component/styles/default.min.css` 
- start supporting `esm` and `cjs` modules. Prepare also compatible with browser umd module.
- supporting almost everything from JSON Schema draft-07 except `dependencies` and `true/false` schemas (due to bug in parser https://github.com/asyncapi/parser-js/issues/232)
- supporting from AsyncAPI 2.0.0 everything except:
  - `id`
  - `operationId`
  - `externalDocs` in operation
  - `discriminator` in AsyncAPI JSON Schema
- `schema` prop can be instance of [`AsyncAPIDocument`](https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L25)
- `config` prop is changed:
  - remove `expand` object due to fact that we want to change expanding functionality
  - move `showErrors` to `show.errors`
  - remove `show.schemas`
  - remove `show.sidebar`
  - rename `show.channels` to `show.operations`
  - remove `pushStateBehaviour` due to fact that we want to change expanding functionality

---

I added `feat:` because I don't think that the current changes (although big) are immediately suitable for 1.0.0 or even 1.0.0-alphaX. We should go further from 0.24.0 and only when we stabilize the component we should go with 1.0.0. Also with 1.0.0-alpha there will be problems with semantic-releases (or not (: ?) @fmvilas @derberg Any thoughts mates?

Thanks for @aeworxet for helping me in this journey! 😄 
 
**Related issue(s)**
See https://github.com/asyncapi/shape-up-process/issues/83
See https://github.com/asyncapi/asyncapi-react/issues/265

PR #321, opened 2021-04-29
ci: add workflows for next branch

magicmatatjahu:
**Description**

As in title. See more [here](https://github.com/asyncapi/asyncapi-react/pull/320#issuecomment-828976524).


.github/workflows/release-wc-next.yml
```diff
@@ -0,0 +1,40 @@
+# This action is a copy of `release-wc-and-playground` workflow for the purpose of the` next` branch only. 
+name: Release Web-component
+
+on: 
+  release:
+    branches:
+      - next
```
derberg, line 7: is there such a thing as a release on a given branch only?

magicmatatjahu: It doesn't work 😅 I changed to `${{ github.event.release.target_commitish }}` in refs.

.github/workflows/release-wc-and-playground.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v2
         with:
-          ref: master
+          ref: ${{ github.event.release.target_commitish }}
```
derberg, line 17: just curious what value is it? can you add a comment?

magicmatatjahu: Added

derberg (commented):
I think you need to do something more to webcomponent workflow. Look at https://www.npmjs.com/package/semantic-release and notice **Published** column. We do not want it to be published as `latest` when released from `next` but as `next`.

We do not use semantic-release here so something has to be changed I think to avoid default

magicmatatjahu:
@derberg You have:

```yaml
      - name: Checkout repository
        uses: actions/checkout@v2
        with:
          # target branch of release. More info https://docs.github.com/en/rest/reference/repos#releases
          ref: ${{ github.event.release.target_commitish }}
      - name: Setup Node.js
        uses: actions/setup-node@v1
        with:
          node-version: 14
      - name: Install latest version of library in the web-component
        run: VERSION=${{github.event.release.tag_name}} npm run install:reactcomp
        working-directory: ./web-component
      - name: Install dependencies
        run: npm install
      - name: Bump version in package.json
        # There is no need to substract "v" from the tag as version script handles it
        # When adding "bump:version" script in package.json, make sure no tags are added by default (--no-git-tag-version) as they are already added by release workflow
        # When adding "bump:version" script in package.json, make sure --allow-same-version is set in case someone forgot and updated package.json manually and we want to avoide this action to fail and raise confusion
        run: VERSION=${{github.event.release.tag_name}} npm run bump:webcomp:version
      - name: Release web-component to NPM
        uses: JS-DevTools/npm-publish@v1
        with:
          token: ${{ secrets.NPM_TOKEN }}
          package: ./web-component/package.json
          access: public
```

So `Bump version in package.json` should handle problem, yes? :)

derberg:
@magicmatatjahu let me say it differently.

We need to make sure that once user runs `npm i @asyncapi/web-component` then `0.23.0` version is installed and to get next they need to do `npm i @asyncapi/web-component@1.0.0-next.1` so in npm **1.0.0-next.1** is not marked as **latest**

magicmatatjahu:
@derberg Changed :)

derberg:
@magicmatatjahu I tried to establish a tunnel connection with Bieruń using putty but I failed, sorry 😄 Maybe easier will be when you actually push your commit to remote 😛

derberg (approved):
lgtm, well done!

PR #326, opened 2021-05-04
ci: fix checkout to next branch

magicmatatjahu:
**Description**

As in title. Fix checkout to `next` branch in `wc-release` and `bump` workflows. Checkout to next when in `release_tag` will contain `next` substring.


derberg (approved):
you use `contains` function in many places. I suggest doing something like:
```
- name: initial step to set variable that will be used in all next steps
   id: release
   run: echo "::set-output name=isNext::$(contains(github.event.release.tag_name, 'next'))"
- name: now I can just use this variable in if to evaluate if next or not
   if: steps.release.outputs.isNext == 'true'
```


not a blocker here, can be a followup, especially that you need to also do it in `.github`. Up to you as I guess you are not 100% sure if this will even work as expected

magicmatatjahu:
@derberg Thanks! First I thought about variable, but I'm afraid if the syntax will work.

derberg:
@magicmatatjahu it worked here https://github.com/asyncapi/asyncapi-react/pull/326/files#diff-ef4f20ad9ac611c67466a8f25e69c3924ed476b297b5dd67f54829c012ea3a00R26-R31 so why not? 😄 just test it in some test project and you know 😄

PR #327, opened 2021-05-04
feat!: unify react-component with html-template

magicmatatjahu:
**Description**

Unify component with with html-template. For more info see https://github.com/asyncapi/shape-up-process/issues/83 and https://github.com/asyncapi/asyncapi-react/issues/265

Changes:
- remove Fiori styling and start using TailwindCSS - theming at the moment won't be possible, we want to add it in the future in simplifier form that currently - see https://github.com/asyncapi/shape-up-process/issues/83 -> Don't support theming section. `@asyncapi/react-component/lib/styles/fiori.css` is deprecated.
- default styles now is in `@asyncapi/react-component/styles/default.min.css` 
- start supporting `esm` and `cjs` modules. Prepare also compatible with browser umd module.
- supporting almost everything from JSON Schema draft-07 except `dependencies` and `true/false` schemas (due to bug in parser https://github.com/asyncapi/parser-js/issues/232)
- supporting from AsyncAPI 2.0.0 everything except:
  - `id`
  - `operationId`
  - `externalDocs` in operation
  - `discriminator` in AsyncAPI JSON Schema
- `schema` prop can be instance of [`AsyncAPIDocument`](https://github.com/asyncapi/parser-js/blob/master/lib/models/asyncapi.js#L25)
- `config` prop is changed:
  - remove `expand` object due to fact that we want to change expanding functionality
  - move `showErrors` to `show.errors`
  - remove `show.schemas`
  - remove `show.sidebar`
  - rename `show.channels` to `show.operations`
  - remove `pushStateBehaviour` due to fact that we want to change expanding functionality

---

I added `feat:` because I don't think that the current changes (although big) are immediately suitable for 1.0.0 or even 1.0.0-alphaX. We should go further from 0.24.0 and only when we stabilize the component we should go with 1.0.0. Also with 1.0.0-alpha there will be problems with semantic-releases (or not (: ?) @fmvilas @derberg Any thoughts mates?

Thanks for @aeworxet for helping me in this journey! 😄 
 
**Related issue(s)**
See https://github.com/asyncapi/shape-up-process/issues/83
See https://github.com/asyncapi/asyncapi-react/issues/265


derberg (approved):
you know you can do empty commit?

magicmatatjahu:
@derberg A man learns all his life 😅

PR #332, opened 2021-05-05
refactor: infer types in proper way and add tests

magicmatatjahu:
**Description**

Infer types in proper way:
- handle true/false schemas,
- infer types from keywords and from `enum`,
- handle empty schema,
- update `bump.yaml` workflow
Add missed tests for helpers (and also fix some bugs).

### DIFF
--- a/library/src/helpers/__tests__/schema.test.ts
+++ b/library/src/helpers/__tests__/schema.test.ts
@@ -1,6 +1,71 @@
 import { SchemaHelpers } from '../schema';
 
+// @ts-ignore
+import Schema from '@asyncapi/parser/lib/models/schema';
+
 describe('SchemaHelpers', () => {
+  describe('.toSchemaType', () => {
+    test('should handle true schemas', () => {
+      const result = SchemaHelpers.toSchemaType(true);
+      expect(result).toEqual(`any`);
+    });
+
+    test('should handle false schemas', () => {
+      const result = SchemaHelpers.toSchemaType(false);
+      expect(result).toEqual(`never`);
+    });
+
+    test('should handle flat types', () => {
+      const schema = new Schema({ type: 'string' });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`string`);
+    });
+
+    test('should handle union types', () => {
+      const schema = new Schema({ type: ['string', 'integer', 'number'] });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`string | integer | number`);
+    });
+
+    test('should infer types', () => {
+      const schema = new Schema({
+        pattern: '^foo',
+        properties: {},
+        items: { multipleOf: 2 },
+      });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`string | array<number> | object`);
+    });
+
+    test('should handle union types in array', () => {
+      const schema = new Schema({ items: { multipleOf: 2, type: 'string' } });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`array<string | number>`);
+    });
+
+    test('should handle tuple types', () => {
+      const schema = new Schema({
+        items: [{ properties: {} }, { pattern: '^foo' }, {}],
+      });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`array<object, string, any>`);
+    });
+
+    test('should handle tuple types', () => {
+      const schema = new Schema({
+        items: [{ properties: {} }, { pattern: '^foo' }, {}],
+      });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`array<object, string, any>`);
+    });
+
+    test('should handle combined types', () => {
+      const schema = new Schema({ type: 'string', oneOf: [] });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`string oneOf`);
+    });
+  });
+
   describe('.prettifyValue', () => {
     test('should handle string', () => {
       const result = SchemaHelpers.prettifyValue('foobar');
--- a/library/src/helpers/__tests__/schema.test.ts
+++ b/library/src/helpers/__tests__/schema.test.ts
@@ -10,6 +10,12 @@ describe('SchemaHelpers', () => {
       expect(result).toEqual(`any`);
     });
 
+    test('should handle empty schema', () => {
+      const schema = new Schema({});
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`any`);
+    });
+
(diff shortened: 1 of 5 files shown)


library/src/helpers/__tests__/schema.test.ts
```diff
@@ -27,4 +120,264 @@ describe('SchemaHelpers', () => {
...
+
+    test('should handle not empty object', () => {
+      const schema = new Schema({ minProperties: 1 });
+      const result = SchemaHelpers.humanizeConstraints(schema);
+      expect(result).toEqual(['non-empty']);
+    });
+  });
+
+  describe('.isExpandable', () => {
+    test('object should be expandable', () => {
```
jonaslagoni, line 241: What is not considered "expandable"? Since all these tests for `isExpandable` all results in true.

library/src/helpers/schema.ts
```diff
@@ -206,34 +209,125 @@ export class SchemaHelpers {
...
+    // object
+    maxProperties: 'object',
+    minProperties: 'object',
+    required: 'object',
+    properties: 'object',
+    patternProperties: 'object',
+    propertyNames: 'object',
+    dependencies: 'object',
+    dependentRequired: 'object',
+    dependentSchemas: 'object',
```
jonaslagoni, line 245: Why does this define keywords not located in JSON Schema?

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -27,4 +120,264 @@ describe('SchemaHelpers', () => {
...
+    });
+
+    test('else should be expandable', () => {
+      const schema = new Schema({ else: {} });
+      const result = SchemaHelpers.isExpandable(schema);
+      expect(result).toEqual(true);
+    });
+  });
+
+  describe('.getCustomExtensions', () => {
```
jonaslagoni, line 296: I would probably add a test displaying that it does not return non-extensions, i.e. if `properties` are defined. Not sure if it helps the coverage since the other test does something similar.

jonaslagoni (commented):
Only had a few questions and suggestions 😄

library/src/helpers/schema.ts
magicmatatjahu: They are, but in 2012-12 draft 😅 Perhaps we should only focus of draft-07, ok I will remove it :)

library/src/helpers/__tests__/schema.test.ts
magicmatatjahu: Primitives without combined schemas :) I will add some tests. Thanks!

library/src/helpers/schema.ts
jonaslagoni: We are only using draft 7 in the spec (at the moment) 🙂

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -27,4 +120,264 @@ describe('SchemaHelpers', () => {
...
+    test('should handle number/integer inclusive range', () => {
+      const schema = new Schema({ minimum: 2, maximum: 5 });
+      const result = SchemaHelpers.humanizeConstraints(schema);
+      expect(result).toEqual(['[ 2 .. 5 ]']);
+    });
+
+    test('should handle number/integer exclusive range', () => {
+      const schema = new Schema({ minimum: 2, exclusiveMaximum: 5 });
+      const result = SchemaHelpers.humanizeConstraints(schema);
+      expect(result).toEqual(['[ 2 .. 5 )']);
```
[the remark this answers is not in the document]
magicmatatjahu: Notation `)` in mathematic means exclusivity, so `[ 2 .. 5 )` means from 2 to 4.9999999 😄 In integer case it means range 2-4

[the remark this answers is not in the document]
jonaslagoni: Ahh! Didn't even notice it was `)` 😆

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -1,6 +1,99 @@
 import { SchemaHelpers } from '../schema';
 
+// @ts-ignore
```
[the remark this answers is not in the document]
magicmatatjahu: We only have types for signature of classes, so if I import some models from root like `@asyncapi/parser` then I cannot create type by `new {model}` - we have only types in the root of `@asyncapi/parser` :)

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -1,6 +1,99 @@
...
+        items: [{ properties: {} }, { pattern: '^foo' }, {}],
+      });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`array<object, string, any>`);
+    });
+
+    test('should handle combined types', () => {
+      const schema = new Schema({ type: 'string', oneOf: [] });
+      const result = SchemaHelpers.toSchemaType(schema);
+      expect(result).toEqual(`string oneOf`);
```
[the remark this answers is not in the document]
magicmatatjahu: It's sound a weird but please consider example:

```js
{
  type: string
  oneOf: [
    {
      minLength: 2,
      maxLength: 5
    },
    {
      minLength: 10,
      maxLength: 15
    }
  ]
}
```

so `string oneOf` probably tells more that `OneOf<String>` when it's not a union type, but single type with different constraints :)

library/src/helpers/schema.ts
magicmatatjahu: Yep, but I like new keywords from schema :D

magicmatatjahu:
@jonaslagoni I added missed tests :)

library/src/helpers/__tests__/schema.test.ts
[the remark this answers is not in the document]
jonaslagoni: Fair enough 🙂

library/src/helpers/schema.ts
jonaslagoni: I mean whether or not you like them shouldn't really matter 😄 If you should not be able to define them in the spec why should they be accounted for in the code?

magicmatatjahu: I've already prepared for the upgrade to draft 2021 😄

library/src/helpers/__tests__/schema.test.ts
magicmatatjahu: Added :)

library/src/helpers/schema.ts
magicmatatjahu: Removed :)

jonaslagoni: Ahh didn't see it. 

But yea, be careful about pre-optimizing these things since we have nothing in the pipeline that changes the payload schema, cause you would have to go in and remove it if we never support it. 😄

jonaslagoni (approved):

magicmatatjahu:
@jonaslagoni Thanks :)

Issue #63, opened 2019-09-11
Support missing fields/objects from spec 2.0.0


magicmatatjahu:
**Description**

At the moment we support a main part of spec in rendering, but there are also parts that we don't, like `bindings`, `properties` from `schema` (min, max, type etc) - https://github.com/asyncapi/asyncapi/blob/master/versions/next/asyncapi.md#properties, externalDocs, tags, correlationID etc.

AC:
- support all missing fields/objects in rendering. With any question about design/styling, please contact with @fmvilas or our designers.

List:
- [X] `properties` from `schema` (min, max, type etc), - https://github.com/asyncapi/asyncapi-react/commit/6e2eeeea3b4431b7e6bec9abd7425d0c6c26a5eb,
- [X] bindings - https://github.com/asyncapi/asyncapi-react/pull/259
- [X] externalDocs - https://github.com/asyncapi/asyncapi-react/pull/327
- [X] tags - https://github.com/asyncapi/asyncapi-react/pull/327
- [X] correlationID - https://github.com/asyncapi/asyncapi-react/pull/327

Updated list (4 May 2021):
- [x] dependencies in JSON Schema visualiser - https://github.com/asyncapi/asyncapi-react/pull/334
- [x] true/false JSON Schemas - https://github.com/asyncapi/asyncapi-react/pull/367
- [x] id https://github.com/asyncapi/asyncapi-react/pull/345
- [x] operationId https://github.com/asyncapi/asyncapi-react/pull/345
- [x] externalDocs in operation https://github.com/asyncapi/asyncapi-react/pull/345
- [ ] discriminator in AsyncAPI JSON Schema - it will be handled in https://github.com/asyncapi/html-template/issues/68

fmvilas:
FYI, protocolInfo = bindings. Bindings is the new name.

hesyar:
We need this feature too.

GreenRover:
+1

derberg:
@hesyar this component was donated by the team behind this project https://kyma-project.io/ and they are main maintainers atm. Afaik, they are super busy atm at least until May. Would it be something you could contribute? AsyncAPI core team is more focused on generator and parser to have 1.0 release for both

hesyar:
Hi @derberg thanks for your reply. For what I can do now is to update the parser version. Furthermore I'll try to contribute more if I find any time. Thanks, I'll let you know as soon as I've the new version.

derberg:
[...]

PR #334, opened 2021-05-06
chore: render schema dependencies

magicmatatjahu:
**Description**

Render schema dependencies in `Schema component`. For example, schema like this:

```yaml
    dependenciesObject:
      type: object
      properties:
        name:
          type: string
        credit_card:
          type: integer
        billing_address:
          type: string
        schema_dependency:
          type: string
      required:
        - name
      dependencies:
        credit_card: [billing_address]
        billing_address: [credit_card]
        schema_dependency:
          properties:
            additionalProp:
              type: string
          required:
            - name
```

will be rendered as:

image

**Related issue(s)**
Part of https://github.com/asyncapi/asyncapi-react/issues/63
Blocked by https://github.com/asyncapi/asyncapi-react/pull/332


library/src/helpers/__tests__/schema.test.ts
```diff
@@ -423,4 +397,133 @@ describe('SchemaHelpers', () => {
...
+  });
+
+  describe('.getDependentRequired', () => {
+    test('should return undefined when dependencies is not defined', () => {
+      const schema = new Schema({
+        properties: {
+          foo: { type: 'string' },
+          bar: { type: 'string' },
+          zor: { type: 'string' },
+        },
```
jonaslagoni, line 434: You have a lot of tests that are not minimalistic, is there a reason why you use more complicated inputs to test the basics?

library/src/helpers/schema.ts
```diff
@@ -200,6 +182,73 @@ export class SchemaHelpers {
     return new SchemaClass(json);
   }
 
+  static getCustomExtensions(value: any) {
```
jonaslagoni, line 185: Please provide at least jsdoc for the function 🙂

library/src/helpers/schema.ts
```diff
@@ -200,6 +182,73 @@ export class SchemaHelpers {
...
+        ) {
+          obj[extName] = ext;
+        }
+        return obj;
+      },
+      {},
+    );
+  }
+
+  static getDependentRequired(
```
jonaslagoni, line 203: Please provide at least jsdoc for the function 🙂

library/src/helpers/schema.ts
```diff
@@ -200,6 +182,73 @@ export class SchemaHelpers {
...
+
+    for (const [prop, array] of Object.entries(dependencies)) {
+      if (Array.isArray(array) && array.includes(propertyName)) {
+        dependentRequired.push(prop);
+      }
+    }
+    return dependentRequired.length ? dependentRequired : undefined;
+  }
+
+  static getDependentSchemas(schema: Schema): Schema | undefined {
```
jonaslagoni, line 221: Please provide at least jsdoc for the function 🙂

jonaslagoni (changes requested):
Looks great, only a couple of comments 👍 

Btw, how do you handle nested schema dependencies? Example
```
{
  "type": "object",
  "properties": {
    "name": { "type": "string" },
    "credit_card": { "type": "number" }
  },
  "dependencies": {
    "credit_card": {
        "properties": {
          "billing_address": { "type": "string" },
          "billing_address2": { "type": "string" }
        },
        "required": ["billing_address"],
        "dependencies": {
          "billing_address": {
            "properties": {
              "billing_address3": { "type": "string" }
            },
            "required": ["billing_address3"]
          }
        }
    }
  }
}
```

library/src/helpers/__tests__/schema.test.ts
magicmatatjahu: I don't see any pluses by testing such functions with simple, a little complicated and complicated examples, but I will add something

library/src/helpers/schema.ts
magicmatatjahu: I don't put any JSDocs for these functions and also for another, because we only export AsyncAPI component from this package, so this functionality is private and the moment only used in one place :)

magicmatatjahu:
@jonaslagoni I added more simple/complex test cases :)

library/src/helpers/schema.ts
jonaslagoni: Done see it as only for the external user, but also for us maintainers, so we can quickly get an overview over what the functions do.

magicmatatjahu:
@jonaslagoni Your example is rendered as :)

image

library/src/helpers/__tests__/schema.test.ts
jonaslagoni: My point is, why do you need 3 properties in the schema to test this functionality 🙂 ?

jonaslagoni:
> @jonaslagoni Your example is rendered as :)

Where is the information about `billing_address3`? 🤔

magicmatatjahu:
@jonaslagoni Here (I didn't toggle last item) :D 
image

jonaslagoni:
> @jonaslagoni Here (I didn't toggle last item) :D

🤦 Great!

library/src/helpers/__tests__/schema.test.ts
magicmatatjahu: To check if someone put multiple keys in `dependencies` array :) You can put input as in tests and it's still valid.

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -423,4 +397,205 @@ describe('SchemaHelpers', () => {
...
+            required: ['billing_address'],
+          },
+          billing_address: {
+            properties: {
+              account: { type: 'string' },
+            },
+            required: ['account'],
+          },
+        },
+      };
```
jonaslagoni, line 575: Maybe it is better the complex case is for nested dependencies? 🤔 Or does this test covers lines that are not tested by the simple one?

library/src/helpers/__tests__/schema.test.ts
jonaslagoni: True I see it.

library/src/helpers/__tests__/schema.test.ts
```diff
@@ -423,4 +397,205 @@ describe('SchemaHelpers', () => {
...
+            },
+            required: ['billing_address'],
+          },
+          billing_address: {
+            properties: {
+              account: { type: 'string' },
+            },
+            required: ['account'],
+          },
+        },
```
jonaslagoni, line 574: Maybe it is better the complex case is for nested dependencies? 🤔 Or does this test covers lines that are not tested by the simple one?

magicmatatjahu: The function which handle `dependencies` works only in this same level as normal `object`, it means that if you put nested `dependencies` so then it is treated as separate schema which is rendered and then function for dependencies again is performed for these second `dependencies`. Component here works in recursive way, it checks that we have another JSON Schema and then it tries to render properties/dependencies etc as another, more deep scheams.

jonaslagoni: Fair enough 🙂

jonaslagoni (approved):
👍

magicmatatjahu:
@jonaslagoni Thanks! :)

PR #345, opened 2021-05-18
feat: render missed parts for operations and render spec id

magicmatatjahu:
**Description**

Render missed parts for operation:
- `operationId`
- `externalDocs`

Render `id` of spec.

ID:
image

Operation:
image

**Related issue(s)**
Part of https://github.com/asyncapi/asyncapi-react/issues/63

### DIFF
--- a/playground/src/specs/streetlights.ts
+++ b/playground/src/specs/streetlights.ts
@@ -102,6 +102,9 @@ channels:
     subscribe:
       summary: Receive information about environmental lighting conditions of a particular streetlight.
       operationId: receiveLightMeasurement
+      externalDocs:
+        description: Find more info here
+        url: https://example.com
       traits:
         - $ref: '#/components/operationTraits/kafka'
       message:
@@ -130,6 +133,9 @@ channels:
         $ref: '#/components/parameters/streetlightId'
     publish:
       operationId: turnOn
+      externalDocs:
+        description: Find more info here
+        url: https://example.com
       traits:
         - $ref: '#/components/operationTraits/kafka'
       message:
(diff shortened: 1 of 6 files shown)


playground/src/specs/streetlights.ts
```diff
@@ -140,7 +146,6 @@ channels:
       streetlightId:
         $ref: '#/components/parameters/streetlightId'
     publish:
-      operationId: turnOff
```
jonaslagoni, line 143: Why remove the operationId? 🤔

jonaslagoni (commented):
Only 1 comment, but, where are the tests? 🧐

playground/src/specs/streetlights.ts
magicmatatjahu: To check if without `operationId` everything is ok :)

magicmatatjahu:
> Only 1 comment, but, where are the tests? 🧐

We haven't integrated test framework for React.

jonaslagoni:
> We haven't integrated test framework for React.

I see jest is already and a few tests have already been added, shouldn't take you long to test the new output i.e. find operation id and externalDocs and the document id. But of course, features take a little longer to create.

You have to start eventually, so why not this PR? 🤔

magicmatatjahu:
>I see jest is already and a few tests have already been added, shouldn't take you long to test the new output i.e. find operation id and externalDocs and the document id. But of course, features take a little longer to create.

Testing normal TS function isn't this same as testing React, it isn't testing html as string and make some regex. Also we will rewrite component again in near future, when we will have a designer and therefore I don't want to test every part of a component.

jonaslagoni:
> Testing normal TS function isn't this same as testing React, it doesn't like testing html as string and make some regex. Also we will rewrite component again in near future, when we will have a designer and therefore I don't want to test every part of a component.

Okay 🙂

jonaslagoni (approved):

Issue #209, opened 2020-12-13
examples for string properties with format `time` show "string"


c-pius:
**Description**

Example values for `{type: "string", format: "time"}` objects return `"string"` instead of a time value.

**Expected result**

Examples for time formatted strings should be something like `"14:15:22"` 

**Actual result**

time property reads `"string"`

![image](https://user-images.githubusercontent.com/22994291/102018765-b3ac2680-3d6f-11eb-867b-ae40034ee3f4.png)

**Steps to reproduce**

Add the following to an arbitrary payload

```json
"some-time": {
    "type": "string",
    "format": "time"
}
```

**Troubleshooting**

I assume this is an inherited problem from the sample function of openapi-sampler
https://github.com/asyncapi/asyncapi-react/blob/36d3a21dabedb0e8c74876b7626f9a8045901ea3/library/src/helpers/generateExampleSchema.ts#L6

magicmatatjahu:
@c-pius Hello! Yes, this is a problem with mentioned 3rd package. However, AsyncAPI supports this same formats like in OpenAPI https://github.com/asyncapi/asyncapi/blob/master/versions/2.0.0/asyncapi.md#dataTypeFormat Also in OpenAPI docs https://swagger.io/docs/specification/data-models/data-types/#string you can see that it supports `date` and `date-time` without mentioning `time`, so `openapi-sampler` package doesn't support it. `format` field is an open value, but probably tools go only with defined format in official docs. Only what we can do is make contribution to `openapi-sampler` package, and the bump it in our component, because issue is not related to component itself - making example for `time` inside component isn't an option.

magicmatatjahu:
@c-pius I created issue for that https://github.com/Redocly/openapi-sampler/issues/120

c-pius:
great, thanks for reaching out to those folks! Actually, I must admit that I wasn't aware that `time` format is not defined as part of OpenAPI 3

derberg:
@c-pius yeah, OpenAPI 3 is actually not 100% JSON Schema. I think they want to fix it with OpenAPI 3.1 and `time` format came out in JSON Schema Draft 7 so "after" OpenAPI 3

derberg:
@magicmatatjahu have a look at my comment, because of what I wrote, they might not agree to add support for `time`. Let's give Roman a few days to answer there, if he won't find the time, I'll try to catch him on one special Slack workspace 😄 

[...]

PR #352, opened 2021-05-25
ci: publish playground from next branch

magicmatatjahu:
**Description**

Changes proposed in this pull request:

- publish playground from `next` branch - update appropriate workflow for that
- improve showing errors from parser: break the words

Errors:
image

**Related issue(s)**
Part of https://github.com/asyncapi/asyncapi-react/issues/300

PR #355, opened 2021-05-26
feat: update openapi-sampler to 1.0.0 for better examples generation

magicmatatjahu:
**Description**

Update `openapi-sampler` package to `1.0.0`. Support for more type formats and JSON Schema 7 driven by support for OpenAPI 3.1.
Changes: https://github.com/Redocly/openapi-sampler/pull/122

Displaying `time` format:

image

**Related issue(s)**
Part of https://github.com/asyncapi/asyncapi-react/issues/209

### DIFF
--- a/library/package-lock.json
+++ b/library/package-lock.json
@@ -685,6 +685,11 @@
         "jest-diff": "^24.3.0"
       }
     },
+    "@types/json-schema": {
+      "version": "7.0.7",
+      "resolved": "https://registry.npmjs.org/@types/json-schema/-/json-schema-7.0.7.tgz",
+      "integrity": "sha512-cxWFQVseBm6O9Gbw1IWb8r6OS4OhSt3hPZLkFApLjM8TEXROBuQGLAH2i2gZpcXdLBIrpXuTDhH7Vbm1iXmNGA=="
+    },
     "@types/marked": {
       "version": "2.0.2",
       "resolved": "https://registry.npmjs.org/@types/marked/-/marked-2.0.2.tgz",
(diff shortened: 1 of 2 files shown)


derberg:
@magicmatatjahu I took the liberty to extend description a bit and the PR title too.

You need to change to merge into master and not next

magicmatatjahu:
@derberg Thanks! I described why it must be done in two separate PRs here -> https://github.com/asyncapi/asyncapi-react/pull/352#issuecomment-848633240 It's related to changes in package-lock between master and next branches. If I will make cherrypick from master to next then I will have to fixing rebase issues.

derberg:
@magicmatatjahu sorry but I didn't get it. what is the problem to put it into master, then merge master into next? even if there are differences with conflicts, then just resolve by saying what is in next is priority, then `npm install` to add updated openapi-sampler to package-lock. Or am I missing something?

magicmatatjahu:
Then you will have two different commits if you will change something from cherry picking, yes? It will only complicate topic, really.

derberg:
Do you mean cherry-picking as a concept or as a git functionality? why not just do a simple merge of latest master into next? I don't think there is anything in the master that cannot be pulled into next, right? this means normal PR, with as many commits you want and then squashing and merging into `next` as a single commit, right?

derberg:
🤦🏼 I think now I get it, of course you need to do cherry pick, you are right that 2 separate PRs to next and master are better. Awaiting for a PR to master then

magicmatatjahu:
> why not just do a simple merge of latest master into next? I don't think there is anything in the master that cannot be pulled into next, right?

No, there are a lot of changes between master and next branches. As I remember some dependencies also was removed from next branches in unification time. Really, don't complicated topic in this area. We will have to merge master to next in some time, when we will decide to go finally with `1.0.0`, but it's not worth the trouble for just one package.

derberg (approved):

magicmatatjahu:
@derberg Thanks a lot for patience ❤️

derberg:
@magicmatatjahu remember about it once we see each other in a bar 🍻 😆

Issue #351, opened 2021-05-24
Recursive schema is not rendering


kchaber:
**Description**

Schema which contains some recursive items does not render and the console logs present some StackOverflow exceptions.

**Expected result**

Schema with recursive items should render correctly.

**Steps to reproduce**

To reproduce this issue you can use the following schema where `RecursiveItem` schema contains list of `subItems` of type `RecursiveItem`.

```
asyncapi: '2.0.0'
info:
  title: Recursive API
  version: '1.0.0'

channels:
  recursive-items:
    subscribe:
      message:
        $ref: '#/components/messages/RecursiveItem'

components:
  messages:
    RecursiveItem:
      payload:
        $ref: "#/components/schemas/RecursiveItem"

  schemas:
    RecursiveItem:
      type: object
      properties:
        name:
          type: string
        subItems:
          type: array
          items:
            $ref: '#/components/schemas/RecursiveItem'
```

Tested on both versions: `0.23.1` and `1.0.0-next-3`

magicmatatjahu:
@kchaber Hello! Thanks for reporting the issue! We know about this bug, for more info check https://github.com/asyncapi/parser-js/issues/293 If we will support the recursive schemas in every place in spec (at the moment we only support in properties) then supporting in component will be very easy.

derberg:
@magicmatatjahu what do you mean we support it in properties only? arrays too. This file works well with HTML template before migration to react

magicmatatjahu:
Yes, but as I described in this issue https://github.com/asyncapi/parser-js/issues/293 (with example) circular refs in items aren't handled properly, so supporting in component is disabled. In most cases this will result in a wrong rendered schema.

derberg:
not sure what you mean by `most cases` because so far it was discovered only with your only case and not even with real schema but a dummy schema for testing only 😄 

I'm not saying the bug is not there, it is there and your example shows it pretty nice, I totally agree with you.

On the other hand though, `next` react component was supposed to not introduce any regression so your approach here is not the best in my opinion. 

Technically what's the problem to support current parser functionality? I mean why not enabling it now and support `most cases` 😄 later it is just a matter of upgrading to latest parser and more cases will be supported, right?

magicmatatjahu:
@kchaber Issue is fixed in `1.0.0-next.5`. We also published new playground which has included fixing. https://asyncapi.github.io/asyncapi-react/ Let us know if someone is broken :) Your schema is rendered as below:

[...]

PR #356, opened 2021-05-26
fix: rendering the circular items

magicmatatjahu:
**Description**

Rendering the circular items.

image

**Related issue(s)**
Fixes https://github.com/asyncapi/asyncapi-react/issues/351

### DIFF
--- a/library/src/components/Schema.tsx
+++ b/library/src/components/Schema.tsx
@@ -52,8 +52,12 @@ export const Schema: React.FunctionComponent<Props> = ({
 
   let uid = schema.uid();
 
-  // checking uid for circular items
-  // after fixing https://github.com/asyncapi/parser-js/issues/293 statement should be removed
+  /**
+   * checking uid for circular items
+   * after fixing https://github.com/asyncapi/parser-js/issues/293 statement should be removed
+   * `x-parser-circular` extension is added to every schema which has circular `items` field,
+   * so we must check that `items` is schema (not array of schemas) and infer UID of schema to display which schema is circular (by the name of schema)
+   */
   if (
     isCircular &&
     !uid &&
--- a/library/src/components/Schema.tsx
+++ b/library/src/components/Schema.tsx
@@ -38,16 +38,30 @@ export const Schema: React.FunctionComponent<Props> = ({
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


library/src/components/Schema.tsx
```diff
@@ -38,17 +38,31 @@ export const Schema: React.FunctionComponent<Props> = ({
...
+
+  let uid = schema.uid();
+
+  // checking uid for circular items
+  // after fixing https://github.com/asyncapi/parser-js/issues/293 statement should be removed
+  if (
+    isCircular &&
+    !uid &&
+    schema.items() &&
+    typeof (schema.items() as SchemaType).uid === 'function'
```
derberg, line 61: how come uid could be a function?

magicmatatjahu: Items can be also array of schemas, like tuple type, so for that I used `typeof (schema.items() as SchemaType).uid === 'function'` to check that I operate on schema, not on array of schemas.

magicmatatjahu: Also `uid` is used here to check what exactly schema is circular

derberg: > Items can be also array of schemas, like tuple type, so for that I used typeof (schema.items() as SchemaType).uid === 'function' to check that I operate on schema, not on array of schemas.

this is not that straight forward, can you add a comment to code please 🙏🏼 

> Also uid is used here to check what exactly schema is circular

I don't know what you mean. Tbh I don't remember how it is handled on the parser side. You mean that circular schema has no `uid`?

magicmatatjahu: > You mean that circular schema has no uid?

Circular schemas have uid, but problem with `items` is that `uid` is added to schema in `items` field but `x-parser-circular` info is added to parent schema, which has `items` field 😅 So I must do this "tricky" think here - I have info that schema has circular ref in items, but I must read uid from `items()` function

derberg: 🤯 
this one definitely deserves a code comment, way to deep knowledge about the parser needed to get the code 😄

magicmatatjahu: Added :)

derberg (approved):

PR #361, opened 2021-05-27
feat: highlight code blocks in markdown

magicmatatjahu:
**Description**

Highlight code blocks in markdown:
- for highlights use `highlight.js` and load styles only for 4 languages: json, yaml and bash due to fact that highlight with all languages (about 138) weight about 800-900kbs - highlights with 3 languages add only about additional 30kb to package. Every other language that mentioned will use as fallback the normal `text` styling.
- export highlight from package to dynamic import language configs. 
- add docs.
- highlight code blocks in markdown in `descriptions` fields etc, in every places where user can define markdown in spec
- highlight examples in this same way and styles like normal markdown - remove old `formatJsonHighlight` solution
- add tests
- update `streetlights.yaml` example with code blocks in few places 
- 🤦🏼 remove leftover of `fiori.css`... 
- changes are used in https://github.com/asyncapi/html-template/pull/210

image

image

image

**Related issue(s)**
Resolves https://github.com/asyncapi/asyncapi-react/issues/354


docs/features/highlight.md
```diff
@@ -0,0 +1,23 @@
+# Highlight markdown's code blocks
+
+## Overview
+
+The component uses internally to highlight the markdown's code blocks an instance of [`highlight.js`](https://highlightjs.org/). To reduce the size of the component, it only uses the configuration for 3 languages: `json`, `yaml` and `bash`. The rest are rendered as plain text. However, the package provides the instance of the [`highlight.js`](https://highlightjs.org/) outside the package with which you can, for example, add a custom language configuration.
```
jonaslagoni, line 5:
```suggestion
The component internally uses [`highlight.js`](https://highlightjs.org/) to highlight the markdown code blocks. To reduce the size of the component, it only uses the configuration for 3 languages: `json`, `yaml` and `bash`. The rest are rendered as plain text. However, the package provides the instance of the [`highlight.js`](https://highlightjs.org/) outside the package with which you can, for example, add a custom language configuration.
```

jonaslagoni (commented):
Only had one comment 👍

docs/features/highlight.md
```diff
@@ -0,0 +1,23 @@
+# Highlight markdown's code blocks
+
+## Overview
+
+The component internally uses [`highlight.js`](https://highlightjs.org/) to highlight the markdown code blocks. To reduce the size of the component, it only uses the configuration for 3 languages: `json`, `yaml` and `bash`. The rest are rendered as plain text. However, the package provides the instance of the [`highlight.js`](https://highlightjs.org/) outside the package with which you can, for example, add a custom language configuration.
```
jonaslagoni, line 5: 🤔 Why is it you include those 3 specific languages?

magicmatatjahu: If you have description in md, you probably will use (at least 90% 😄 ) `json`, `yaml` or `bash` language to describe config for server in json/yaml, maybe some bash script to make curl etc - they are the most used languages ​​when it comes to describing something so they are added by default. Here I am based on the probability of use. Additionally, `json` is used for rendering examples for messages and these three languages and they don't add that much kbs to the package.

jonaslagoni: Okay fair enough 🙂

jonaslagoni (approved):
👍

PR #367, opened 2021-06-14
fix: support true/false JSON schemas

magicmatatjahu:
**Description**

Support `true/false` schemas in `Schema` component:
- render `Any` type for `true` and `{}` schemas
- render `Never` type for `false` and `{ not: {} }` schemas
- infer types only for `enum` and `const`
- improve rendering of tuple type and array type
- add tests for it
- update `markedjs` package to `2.1.1` 
- create `tsconfig.json` for handling types in tests 

Why infer types only for `enum` and `const` types? I made a bug with inferring types from keywords like properties, items etc... If someone will write this schema:

```js
{
  properties: {
    foo: { type: "string" }
  },
}
```

it doesn't mean that type for this schema is only `object` - it's not the truth, because this schema means that every type is allowed but only object is restricted with `properties` shape etc, so for this approach we will return the `Restricted Any` type. Any better ideas how to name it? 

Example:
image

Rendering the tuple:
image

Rendering of array:
image

**Related issue(s)**
See more https://github.com/asyncapi/parser-js/issues/232
Fixes https://github.com/asyncapi/asyncapi-react/issues/377


derberg (commented):
LGTM except for this `any` and `restricted any` strings that are used in different places, tests, and main logic, and now they are "hardcoded". 

I'm not sure they should be added to `jsonSchemaKeywordTypes`, maybe there should me a separate type for these ANY types? 🤔

magicmatatjahu:
@derberg I moved all custom types like `any`, `never`, `restricted any` and `unknown` to `SchemaCustomTypes` enum and also add code to handle the case when `items` with `array` type is falsy. Btw. Thanks very much for the review!

derberg (approved):
:shipit:

Issue #369, opened 2021-06-15
Add to all TailwindCSS classes the prefix


magicmatatjahu:
**Description**

In the next version of component we use the TailwindCSS for styling. We don't use the prefixes for TailwindCSS classes used inside the component. It can ends with duplicated classes with other properties than ours in some application - in other words, users can attach to application existing TailwindCSS stylings based on "normal" TailwindCSS class names, so they can override our styles or our styles can override other.

This approach is used in several project like stoplight (they use the `sl` prefix) etc.

We should use the `prefix` options for TailwindCSS https://tailwindcss.com/docs/configuration#prefix but the question is, what prefix? `ai` as a shortcut of `asyncapi` or maybe another?

fmvilas:
I don't really see what's the problem of using the regular TailwindCSS classes. Mind explaining a bit better?

magicmatatjahu:
@fmvilas

> in other words, users can attach to application existing TailwindCSS stylings based on "normal" TailwindCSS class names, so they can override our styles or our styles can override other.

Also in TailwindCSS docs for `prefix`:

> The prefix option allows you to add a custom prefix to all of Tailwind’s generated utility classes. This can be really useful when layering Tailwind on top of existing CSS where there might be naming conflicts

People can have in application own styles based on custom TailwindCSS config, other that us, and it could ends with overriding the styles in their app, or in our component.

fmvilas:
I see. Thanks for explaining!

As a prefix, it's common to see people using `a2s` as **A**sync**A**pi **S**pecification. Another suggestion would be `arc` as in **A**syncapi **R**eact **C**omponent.

magicmatatjahu:
No problem, I very like the `a2s` :) Thanks!

derberg:
`arc` seems better as related to the component itself

derberg:
yeah, forgot about this `ui` idea. Then `aui` 😉

magicmatatjahu:
Fixed by https://github.com/asyncapi/asyncapi-react/pull/493

PR #374, opened 2021-06-17
feat: add script to generate the theme

magicmatatjahu:
**Description**

Add script to generate the theme:
- create the `tools` folder in which we will have the helper scripts/tools for component, like theme/styles generator
- script creates the two css files, with reset css `{name}.reset.css` and without reset css `{name}.css` - https://github.com/asyncapi/asyncapi-react/issues/368
- script takes two parameters, first is a path to output css file like `./output/index.css` and the second parameter (optional) is the path to the custom configuration for TailwindCSS. We deep merge the default configuration with custom configuration.
- add `theming.md` documentation. 

We can use the script like 

```sh
node ./library/tools/styles/script.js ./output/index.css
// or
node ./library/tools/styles/script.js ./output/index.css ./custom.config.js
```

In the future we can move this script to our cli with `generate theme` commands like:

```bash
asyncapi generate theme --config=... --output=...
```

Is it a good idea? Feel free to comment!

Problems I found:
- dependencies... at the moment we can only add all needed dependencies to generate the styles like `tailwindcss`, `postcss` and their plugins to `dependencies` array, then people will able to generate the styles using script in their project like: `node node_modules/@asyncapi/react-component/tools/styles/script.js ...`, of course we will end up with bigger package, but the component itself in application won't change the weight, because these dependencies are a pure devDependencies that aren't used in component directly, but only in `tools/styles` folder.

**Related issue(s)**
Resolves https://github.com/asyncapi/asyncapi-react/issues/370
Resolves https://github.com/asyncapi/asyncapi-react/issues/368

PR #381, opened 2021-06-22
feat: add prefix for css classes

magicmatatjahu:
**Description**

Add to all TailwindCSS classes the `aui-` prefix. Why? More info is in Resolves https://github.com/asyncapi/asyncapi-react/issues/369:

> In the next version of component we use the TailwindCSS for styling. We don't use the prefixes for TailwindCSS classes used inside the component. It can ends with duplicated classes with other properties than ours in some application - in other words, users can attach to application existing TailwindCSS stylings based on "normal" TailwindCSS class names, so they can override our styles or our styles can override other.

**Related issue(s)**
Resolves https://github.com/asyncapi/asyncapi-react/issues/369
Blocked by https://github.com/asyncapi/asyncapi-react/pull/374

### DIFF
--- a/library/src/components/CollapseButton.tsx
+++ b/library/src/components/CollapseButton.tsx
@@ -9,7 +9,7 @@ export const CollapseButton: React.FunctionComponent<Props> = ({
   children,
   ...rest
 }) => (
-  <button {...rest} className={`focus:outline-none ${rest.className}`}>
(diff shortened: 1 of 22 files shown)


magicmatatjahu:
Fixed by https://github.com/asyncapi/asyncapi-react/pull/493

PR #387, opened 2021-07-02
fix: update openapi-sampler to 1.1.0

magicmatatjahu:
**Description**

- Update `openapi-sampler` to `1.1.0` to fix the https://github.com/asyncapi/html-template/issues/195 issue
- Update the `streetlight` example

Generated example from provided schema in mentioned issue:

image

**Related issue(s)**
Fixes https://github.com/asyncapi/html-template/issues/195

### DIFF
--- a/library/package-lock.json
+++ b/library/package-lock.json
@@ -7186,9 +7186,9 @@
       "dev": true
     },
     "openapi-sampler": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/openapi-sampler/-/openapi-sampler-1.0.0.tgz",
-      "integrity": "sha512-HysKj4ZuLk0RpZkopao5SIupUX8LMOTsEDTw9dSzcRv6BBW6Ep1IdbKwYsCrYM9tnw4VZtebR/N5sJHY6qqRew==",
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/openapi-sampler/-/openapi-sampler-1.1.0.tgz",
+      "integrity": "sha512-/LhZYKNBWphLEpbAG5BdpBZbIbmLgC4vTiTj8N/MV0LF9ptmKOiJ2nETVlacNjXHt7iqDgZDELJCIoZ3q5ZG6A==",
       "requires": {
         "@types/json-schema": "^7.0.7",
         "json-pointer": "^0.6.1"
--- a/library/package.json
+++ b/library/package.json
@@ -71,7 +71,7 @@
     "highlight.js": "^10.7.2",
     "isomorphic-dompurify": "^0.13.0",
     "marked": "^2.1.1",
-    "openapi-sampler": "^1.0.0",
+    "openapi-sampler": "^1.1.0",
     "use-resize-observer": "^7.0.0"
   },
   "devDependencies": {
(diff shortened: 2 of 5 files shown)


derberg (commented):
- why did you remove asyncapi message.example in favour of json schema example?
- will you update dummy example too?

magicmatatjahu:
@derberg

> why did you remove asyncapi message.example in favour of json schema example?

To see the output from `openapi-sampler`

> will you update dummy example too?

Spec from this repo has probably been changed for almost 20 commits and has nothing similar with `dummy.yaml` 😄  I can remove changes, I only update spec with schemas from https://github.com/asyncapi/html-template/issues/195 to see, as I wrote, the output from `openapi-sampler`.

derberg:
> but to check that tooling which we are using in component works properly.

this is exactly the purpose of dummy example, to have as many different options from AsyncAPI as possible to test components.

the problem I see that before introducing react in html-template, the html-template covered dummy example. Now it is validated against this super custom streetlight example. We had a discussion in the past that long term we should modify street light only in a way it makes sense as an example (this is why we have now 2 official examples, one for kafka and one for mqtt).  Thus my questions, what is your plan here 😄 

> To see the output from openapi-sampler

you mean only for the review?

magicmatatjahu:
> you mean only for the review?

Not only, also for preview for developing.

> what is your plan here 😄

Leave the new schemas 😄  I remember about discussion, but what? Modify the dummy.yaml to cover the problem mentioned in https://github.com/asyncapi/html-template/issues/195? About dummy, if we really want to treat dummy example as example with all features which are included in AsyncAPI spec, we should move the example from this https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml path to probably https://github.com/asyncapi/spec/tree/master/examples here.

derberg:
> Not only, also for preview for developing.

how will you then preview message.examples? if there render properly, as a list if more than 1 + no label saying that they are generated?

> About dummy, if we really want to treat dummy example as example with all features which are included in AsyncAPI spec, we should move the example from this https://github.com/asyncapi/generator/blob/master/test/docs/dummy.yml path to probably https://github.com/asyncapi/spec/tree/master/examples here

sure, propose a better location than generator, but for sure not spec/examples as this is not an example that users should follow as it is dummy, it mixes many features that would never be used by anyone at any time, so this is not an example for users, but an example that we should test our templates against until TCK can be used for that

magicmatatjahu:
> how will you then preview message.examples? if there render properly, as a list if more than 1 + no label saying that they are generated?

You're right, I will use the schema in another place.

> sure, propose a better location than generator, but for sure not spec/examples as this is not an example that users should follow as it is dummy, it mixes many features that would never be used by anyone at any time, so this is not an example for users, but an example that we should test our templates against until TCK can be used for that

If we wait for TCK it... can stay in generator 😄  or should we move to the parser?

derberg:
we never used it to test parser, this is why it is in generator where it is also used in generator tests

magicmatatjahu:
@derberg So should I move the new schemas to the `dummy.yaml` or remove them from PR?

derberg:
this is just a fix PR so I don't want to complicate your life here too much. I suggest add these new schemas to dummy example in the generator independently, and leave what you have here as it is. As a followup remove streetlight and come up with a way to reuse dummy example here, maybe we need a separate issue first to discuss, but I don't think so.

Basically, I do not want to block this PR. You fixed the schema so we do not loose an example and this is most important for this PR

derberg (approved):

PR #389, opened 2021-07-09
feat: enable rendering of extensions on channel level

derberg:
- enable rendering of extensions on channel item object
- change order of rendering where channel is first before operation as channel is a higher level object
- extended dummy https://github.com/asyncapi/generator/pull/613


magicmatatjahu (approved):
LGTM!

PR #416, opened 2021-08-23
feat: render binding in capsule instead of drop down

### LINKED ISSUE #414: Display protocol on bindings in capsule instead of drop down
The idea is to improve the actual bindings UI to match the initial implementation do in the html-template

**Reasons**

- improve the user experience when reading bindings blocks
- gain one level on rendering (because the protocol part is attached to the bindings title) into the bindings parts

**Attachments**

Here is the previous rendering (available on html-template 0.20) on the left side and the actual rendering on the right side

image

magicmatatjahu:
**Description**

Render binding information as single component: in capsule instead of drop down.

Example:

image

**Related issue(s)**
Resolves #414

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,2 +1,0 @@
-# Disable cognitive complexity in files
-sonar.exclusions=library/src/components/Schema.tsx
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,2 +1,2 @@
 # Disable cognitive complexity in files
-sonar.cpd.exclusions=library/src/components/Schema.tsx
+sonar.exclusions=library/src/components/Schema.tsx
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,2 @@
+# Disable cognitive complexity in files
+sonar.exclusions=library/src/components/Schema.tsx
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,2 @@
+# Disable cognitive complexity in files
+sonar.cpd.exclusions=library/src/components/Schema.tsx
--- a/library/src/components/Bindings.tsx
+++ b/library/src/components/Bindings.tsx
@@ -10,13 +10,28 @@ interface Props {
 }
 
 export const Bindings: React.FunctionComponent<Props> = ({
-  name = 'Bindings',
+  name = 'Binding',
   bindings,
 }) => {
   if (!bindings || !Object.keys(bindings).length) {
     return null;
   }
 
-  const schema = SchemaHelpers.jsonToSchema(bindings);
-  return <Schema schemaName={name} schema={schema} />;
+  const renderedBindings = Object.entries(bindings).map(
+    ([bindingName, binding]) => {
+      const schema = SchemaHelpers.jsonToSchema(binding);
+      const schemaName = (
+        <div className="inline-block text-sm">
+          <span>{name}</span>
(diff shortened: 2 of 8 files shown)


library/src/components/Bindings.tsx
```diff
@@ -10,13 +10,28 @@ interface Props {
 }
 
 export const Bindings: React.FunctionComponent<Props> = ({
-  name = 'Bindings',
+  name = 'Binding',
```
M3lkior, line 13: What about using the `specific information ` suffix as text (validated with @derberg in the original implementation). This provides a more comprehensive label for non-tech which are reading the generated documentation ;

library/src/containers/Operations/Operation.tsx
```diff
@@ -118,7 +118,7 @@ export const Operation: React.FunctionComponent<Props> = ({
 
         {channel.hasBindings() && (
           <div className="mt-2">
-            <Bindings name="Channel Bindings" bindings={channel.bindings()} />
+            <Bindings name="Channel Binding" bindings={channel.bindings()} />
```
M3lkior, line 121:
```suggestion
            <Bindings name="Channel specific information" bindings={channel.bindings()} />
```

library/src/containers/Operations/Operation.tsx
```diff
@@ -127,7 +127,7 @@ export const Operation: React.FunctionComponent<Props> = ({
         {operation.hasBindings() && (
           <div className="mt-2">
             <Bindings
-              name="Operation Bindings"
+              name="Operation Binding"
```
M3lkior, line 130:
```suggestion
              name="Operation specific information"
```

library/src/components/Bindings.tsx
magicmatatjahu: I changed it and it's an example how it looks:

image

M3lkior: very nice 👌

derberg:
what the?

Screenshot 2021-08-23 at 15 18 37

.sonarcloud.properties
```diff
@@ -0,0 +1,2 @@
+# Disable cognitive complexity in files
+sonar.exclusions=library/src/components/Schema.tsx
```
derberg, line 2: I think it excludes all not just complexity, please ignore complexity on a file-level maybe -> https://github.com/asyncapi/parser-js/blob/master/test/testsUtils.js#L23

magicmatatjahu:
Probably due to Ludovic's review.

derberg (commented):
one comment, also why package-lock files had to be updated?

can you also update description of the PR with latest screen shot?

.sonarcloud.properties
magicmatatjahu: I tried to use the `eslint` and it doesn't work. I know about that sonar exclude all rules, but I cannot find example how to exclude only particular rule, I also cannot find the title/id of the complexity rule...

Also related to the package-lock.json - lerna's `bootstrap` removes the redundant modules in the monorepo :) I will revert this.

library/src/components/Schema.tsx
```diff
@@ -1,11 +1,13 @@
+/* eslint-disable sonarjs/cognitive-complexity */
```
derberg, line 1: I think it works only if you put it right before the affected function

magicmatatjahu: Eslint doesn't work in any form, because then we should have errors from `lint` stage, and we don't.

M3lkior:
> @derberg Probably due to Ludovic's review.

Hum .. i just leave single comment ; no review comment

derberg (approved):
I'm wondering if we can have a better PR title, but of course I have no better suggestion out of my head 😆

PR #445, opened 2021-09-29
fix: make fallback for 2.2.0 channel.servers field

magicmatatjahu:
**Description**

More info https://github.com/asyncapi/html-template/issues/264

Make fallback for new in `2.2.0` AsyncAPI version `channel.servers' field - support case when someone will use newest component with old generator, which isn't support `2.2.0` version of AsyncAPI.

**Related issue(s)**
Fixes https://github.com/asyncapi/html-template/issues/264

### DIFF
--- a/library/src/containers/Operations/Operation.tsx
+++ b/library/src/containers/Operations/Operation.tsx
@@ -37,7 +37,7 @@ export const Operation: React.FunctionComponent<Props> = ({
 
   const operationId = operation.id();
   const externalDocs = operation.externalDocs();
-  // check typeof as fallback for older version that `2.2.0`
+  // check typeof as fallback for older version than `2.2.0`
   const servers = typeof channel.servers === 'function' && channel.servers();
 
   const operationSummary = operation.summary();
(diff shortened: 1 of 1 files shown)


derberg:
I don't believe it is the way to go, you should rather mark in template config that template is compatible only with specific generator version

magicmatatjahu:
> I don't believe it is the way to go, you should rather mark in template config that template is compatible only with specific generator version

I agree, but at the moment we have problem with html-template dependencies - even if someone will use the `0.23.5`, template will always install latest version of `react-component` - https://github.com/asyncapi/html-template/blob/master/package.json#L37 Of course we should fix this dependency in the template to have hardcoded version (not latest) but this fix is needed for people which uses the `>0.21.0` version of html-template (where we introduced the react-component) with `generator-action` (which newest versions are broken as you know).

derberg:
@magicmatatjahu I'm missing something, sorry, but 0.23.5 uses old react component https://github.com/asyncapi/html-template/blob/v0.23.5/package.json#L37

magicmatatjahu:
@derberg `^` is important here - it means that you will install newest version of `next` tag, so you will always end with latest version. By this people have problem with `0.23.5` like in this issue https://github.com/asyncapi/html-template/issues/264 Even though they are using an old version of the template, the template still installs the latest react-component. This ends up with the generator using an old version of the parser with no support for the new field, while the component is the latest one and there is an error that the `servers` function for the channel does not exist.

derberg (approved):
right, these are all patches! you're right, sorry

PR #465, opened 2021-11-04
feat: handle circular refs

magicmatatjahu:
**Description**

Handle circular refs:
- use new logic from https://github.com/asyncapi/parser-js/pull/391 PR
- make fallback for older solution


library/src/components/Schema.tsx
```diff
@@ -47,25 +47,38 @@ export const Schema: React.FunctionComponent<Props> = ({
...
   let uid = schema.uid();
 
-  /**
-   * checking uid for circular items
-   * after fixing https://github.com/asyncapi/parser-js/issues/293 statement should be removed
-   * `x-parser-circular` extension is added to every schema which has circular `items` field,
-   * so we must check that `items` is schema (not array of schemas) and infer UID of schema to display which schema is circular (by the name of schema)
-   */
-  if (
+  // ane logic and fallback for older logic for circular references
```
jonaslagoni, line 60: 🤔

playground/src/specs/circular.ts
```diff
@@ -0,0 +1,141 @@
+export const circular = `
```
jonaslagoni, line 1: Where is this used?

jonaslagoni (changes requested):
Well, tests, I can't know if your feature works as expected 🤷 

Small comments.

library/src/components/Schema.tsx
magicmatatjahu: Update comment 😅

playground/src/specs/circular.ts
magicmatatjahu: It's used in playground. If you develop you can choose schema and it's only used to check that circular (very complex) schemas are also supported.

jonaslagoni (approved):
👍

PR #493, opened 2021-12-02
feat: introduce scoped css for default styles

### LINKED ISSUE #369: Add to all TailwindCSS classes the prefix

kaiszybiak:
**Description**

We integrate the `asyncapi` component as a plugin into the [backstage](https://backstage.io/) ecosystem. During the integration of the pre release version containing the switch to tailwind we noticed, that by including the default styles `import "@asyncapi/react-component/styles/default.css";` those styles where globally applied and had an effect on other components, as they are not scoped (see related [PR](https://github.com/backstage/backstage/pull/8088)). 

This PR is inspired by https://github.com/asyncapi/asyncapi-react/pull/381. Our initial idea was to resolve the merge conflicts and go with the CSS prefix solution. (see [branch](https://github.com/asyncapi/asyncapi-react/compare/next...kaiszybiak:next-branches/tailwind-css-prefix?expand=1)). However the tailwind prefix does not affect: 

```css
/* default.css */
@tailwind base;
@tailwind components;
@tailwind utilities;
```
Our second approach was to use tailwind [selector-strategy](https://tailwindcss.com/docs/configuration#selector-strategy): 

```ts
// tailwind.config.js
module.exports = {
  important: '#app',
}
```
This made the refactoring easier, as we only had to add one id to the container and were not forced to change all of the tailwind utility classes manually but it still had the same problem as the prefix solution, that there where still styles applied globally e.g.

```css
/* snippet from @asyncapi/react-component/styles/default.css */

button:focus {
  outline: 1px dotted;
  outline: 5px auto -webkit-focus-ring-color;
}
```
**Solution**

We scope all CSS by using the `postcss` plugin [postcss-scopify](https://github.com/pazams/postcss-scopify). The default styles will only affect the `asyncapi` component. There is a slightly difference between the following screenshots, as the `default.css` doesn't affect the `CodeMirror` component anymore. 

Essentially all styles stay the same only styles applied to `html` and `body` do not have an effect anymore as they are scoped but the scoping class `.aui-root` is applied some levels deeper to the `AsyncApiLayout`. 

```css
/**
Use a more readable tab size (opinionated).
*/

.aui-root html {
  -moz-tab-size: 4;
  -o-tab-size: 4;
     tab-size: 4;
}

/**
1. Correct the line height in all browsers.
2. Prevent adjustments of font size after orientation changes in iOS.
*/

.aui-root html {
  line-height: 1.15; /* 1 */
  -webkit-text-size-adjust: 100%; /* 2 */
}

```

**Before**:
Bildschirmfoto 2021-12-02 um 14 44 15

**After**:
Bildschirmfoto 2021-12-02 um 14 29 42

- Introduce `postcss` plugin [postcss-scopify](https://github.com/pazams/postcss-scopify)
- Add `aui-root` as a scoped class to the `AsyncApiLayout` component
- Add `leading-normal` to match the line-height applied before to the html tag by `@tailwind base`

**Related issue**
Resolves #369
See also #381


magicmatatjahu:
@kaiszybiak Thanks for great description and solution! Yeah, I went with adding to all classes the prefix but your is better, I like it! That I hadn't thought of something similar either 😅  Thanks again, everything looks great :) I also read the discussion in the attached PR in the backstage and don't worry. The `next` version will not be changed much in terms of styles. The only thing we still want to do is fix the bugs that people have reported and we'll release 1.0.0. We waited so long to release `next` because we were waiting for a designer to be hired, but it took us so long that the new styles won't be available until 2.0.0.

Issue #498, opened 2021-12-10
ResizeObserver loop limit exceeded


dschwank:
**Description**
Within the new `next` version we see some uncaught `ResizeObserver loop limit exceeded` exceptions (especially in E2E tests) as soon as the width is smaller than `1280px`. It seems to be related to this snippets here:
- https://github.com/asyncapi/asyncapi-react/blob/next/library/src/containers/AsyncApi/Layout.tsx#L30-L36
- https://github.com/asyncapi/asyncapi-react/blob/next/library/src/containers/AsyncApi/Layout.tsx#L44

We assume that the `ResizeObserver` is not able to deliver all observations within one animation frame due to the nested `setState()`. At least this comment leads to that assumption: https://stackoverflow.com/a/50387233/11471068

**Expected result**

- the AsyncAPI library does not cause any `ResizeObserver loop limit exceeded` exceptions

**Actual result**

- the AsyncAPI causes from time to time `ResizeObserver loop limit exceeded` exceptions

**Steps to reproduce**

1. Start the Playground on the `next` branch with `npm start`
2. Open the playground at `localhost:3000`
3. Open the devTools
4. Add the following listener to console

```js
window.addEventListener('error', (e) => {
  console.log('[UNCAUGHT]', e);
});
```

5. Resize the playground / code section, thus the panel is shown on the right side

Result: Now you should see a `ResizeObserver loop limit exceeded` in your console

**Troubleshooting**

This issue is not critical in the browser, but it causes our Cypress E2E tests to fail. As mitigation we ignore those uncaught exceptions:

```ts
// credits go to https://github.com/quasarframework/quasar/issues/2233#issuecomment-656464466

Cypress.on('uncaught:exception', (err) => {
  if (err.message.includes('ResizeObserver')) {
    // returning false here prevents Cypress from
    // failing the test
    return false;
  }
  return true;
});
```

magicmatatjahu:
@dschwank Hello! Thanks for the issue. I created PR with possible fix (https://github.com/asyncapi/asyncapi-react/pull/499), but I need to reproduce your problem from the upstream version of component to check that everything is good, but tbh my computer is probably to fast and I don't see any exception from `ResizeObserver` 😅

dschwank:
Hey @magicmatatjahu!

Thanks for your response and for waiting so long. I read your message and tested it right after you posted it, but somehow I missed to answer. Sorry for that!

[...]

PR #499, opened 2021-12-13
fix: improve performance of resizeObserver

magicmatatjahu:
**Description**

Improve performance of `resizeObserver`:
- move changing logic of `className`  to the observer.
- use `requestAnimationFrame`
- fix sonar complixity errors

**Related issue(s)**
Fixes: #498

### DIFF
--- a/library/src/containers/AsyncApi/Layout.tsx
+++ b/library/src/containers/AsyncApi/Layout.tsx
@@ -29,15 +29,17 @@ const AsyncApiLayout: React.FunctionComponent<Props> = ({
 
   const { ref } = useResizeObserver<HTMLDivElement>({
     onResize: ({ width }) => {
-      if (width === undefined) {
-        return;
-      }
+      requestAnimationFrame(() => {
+        if (width === undefined) {
+          return;
+        }
 
-      const possibleClassName =
-        width <= 1280 ? 'container:xl' : 'container:base';
-      if (possibleClassName !== observerClassName) {
-        setObserverClassName(possibleClassName);
-      }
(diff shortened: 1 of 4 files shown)


library/src/containers/AsyncApi/Layout.tsx
```diff
@@ -25,12 +25,18 @@ const AsyncApiLayout: React.FunctionComponent<Props> = ({
...
-      if (divWidth !== width) {
-        width && setDivWidth(width);
+      if (width === undefined) {
+        return;
+      }
+
+      const possibleClassName =
+        width <= 1280 ? 'container:xl' : 'container:base';
+      if (possibleClassName !== observerClassName) {
+        setObserverClassName(possibleClassName);
```
dschwank, line 39: Setting the class instead of the width works already much better for me, but I still receive the error message (see [here](https://github.com/asyncapi/asyncapi-react/issues/498#issuecomment-998794808) ).

I only got around it by using a `setTimeout()`, thus it is not in the same cycle anymore.

```suggestion
        setTimeout(() => setObserverClassName(possibleClassName), 0);
```

magicmatatjahu: I wrapped callback with `requestAnimationFrame`, it fixed problem in my local env, so I think that is also a good option, but thanks for the suggestion! :)

derberg:
@magicmatatjahu SonarCloud doesn't like this PR 😉

magicmatatjahu:
@derberg I fixed errors for you 😘

derberg (approved):

PR #502, opened 2022-01-03
fix: render empty extensions and bindings

magicmatatjahu:
**Description**

Render empty extensions and bindings:
- improve `jsonFieldToSchema` function to handle `null` and `undefined` values.
- add unit tests for helper and also for `Bindings` and `Extensions` component

**Related issue(s)**
Resolves https://github.com/asyncapi/studio/issues/221

### DIFF
--- a/library/src/components/Bindings.tsx
+++ b/library/src/components/Bindings.tsx
@@ -29,7 +29,9 @@ export const Bindings: React.FunctionComponent<Props> = ({
         </div>
       );
       return (
-        <Schema schemaName={schemaName} schema={schema} key={bindingName} />
+        schema && (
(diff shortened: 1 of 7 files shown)


paulRbr (approved):
Ok with this solution, but wdyt about “failing early” in the `.map` entries?

Something like:
```tsx
const renderedBindings = Object.entries(bindings).map(
  ([bindingName, binding]) => {
    if (!binding) {
      return null;
    }
    ...
```

magicmatatjahu:
@paulRbr It can be done, but in 99.99% of cases there should be no null/undefined, so I gave this comparison at the end using the `schema && ...` guard :)

derberg (approved):
we need to start generating studio previews on PRs 😅

PR #511, opened 2022-01-12
fix: do not download schemaUrl several times in a short time in the web-component

### LINKED ISSUE #243: `schemaUrl` it sends a few requests to the same resource
Hello guys, one more question: @aeworxet @derberg @magicmatatjahu 
when I use `schemaUrl` it sends a few requests to the same resource
image

Also when I use my own api to get asyncAPI json the following message is shown until I get a response.
image
Can the `asyncapi-component` be rendered once url/json is parsed?

magicmatatjahu:
**Description**

At the moment web-component downloads in a row 4-5 times this same `schemaUrl` (if passed), this PR fix it. The logic is very simple. Inside React class component which is wrapper for native `web-component` we check if new url was changed looking at the last 25 milliseconds ago. This ensures that the changes are still propagated but with some delay and we only have 1 query for the 25 millisecond period.

**Related issue(s)**
Resolves #243


magicmatatjahu:
Sonarcloud throws warnings that two TS assertions aren't needed because they don't change returned type - `This assertion is unnecessary since it does not change the type of the expression`. We need that assertion because webpack (on bundling) throws error that `url` property doesn't exist in the returned value from `retrieveSchemaProp` function, but it can have, so please don't look at that warnings.

derberg:
@magicmatatjahu but it will stay there anyway, don't matter if I close my eyes 😄 

just silent it with proper comment next to it

magicmatatjahu:
@derberg Should be fixed. I tried not to use `any` 😆

derberg (approved):

PR #514, opened 2022-01-13
feat: add collapse button for nested subschemas in the Schema component

### LINKED ISSUE #440: Ability to auto-expand message schema
## Description

If one has a scheme that is too large and deep, then the component will render a very large amount of nesting that one has to manually click through to see the deep schema values. To increase the UX, we should have an additional button next to subschema button - a button to show deep nesting.

## Solution

The changes should be introduced in the [Schema component](https://github.com/asyncapi/asyncapi-react/blob/next/library/src/components/Schema.tsx). If the user presses the extra button, all subschemas should open/close (depending if subschemas are opened/closed).

---

> **NOTE**: Below is old description

@davydog187 commented on [Wed Jan 22 2020](https://github.com/asyncapi/generator/issues/202)

Hello and thank you for this great project!

## The Problem
I have nested payload structures that are 3-4 levels deep, and it is cumbersome to keep clicking to view the schema. 

## Potential Solution
It would be nice if there was an option in the generator that has nested structures auto-expanded, maybe with a parameter to specify the level of nesting that you want to stop auto-expanding. 

I could be interested in helping contribute, but would need some pointing in the right direction :) Thanks!

---

@fmvilas commented on [Wed Feb 05 2020](https://github.com/asyncapi/generator/issues/202#issuecomment-582338627)

Hey @davydog187. Thanks for the suggestion. It would indeed be great. Any help would be much appreciated 🙏

magicmatatjahu:
**Description**

Add collapse button for nested subschemas in the Schema component. Button is next to the normal collapse button. If the user presses the extra button, all subschemas will open/close (depending if subschemas are opened/closed).

**Related issue(s)**
Resolves #440

### DIFF
--- a/library/src/components/CollapseButton.tsx
+++ b/library/src/components/CollapseButton.tsx
@@ -12,16 +12,52 @@ export const CollapseButton: React.FunctionComponent<Props> = ({
   <button {...rest} className={`focus:outline-none ${rest.className}`}>
     {children}
     <svg
-      version="1.1"
-      viewBox="0 0 24 24"
-      x="0"
       xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 20 20"
+      stroke="currentColor"
+      version="1.1"
+      x="0"
       y="0"
+      fill="none"
       {...chevronProps}
-      className={`inline-block align-baseline cursor-pointer -mb-1 w-5 transform transition-transform duration-150 ease-linear ${chevronProps?.className ||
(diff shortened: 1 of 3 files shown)


magicmatatjahu:
@mcturco I know that you're not a maintainer of that branch but could you check only design of that new button (the right one) and UX and write something about it, do you accept/reject?

magicmatatjahu:
It will be handled in https://github.com/asyncapi/asyncapi-react/pull/604

Issue #549, opened 2022-02-16
the pattern field is displayed as lowercase that is not appropriate for pattern like [A-Z]{3}


M3lkior:
**Description**

the display of the pattern field is lower cased that is it not appropriate when using a regex like `[A-Z]{3}` that is displayed as  `[a-z]{3}`

**Expected result**

Display the pattern like it is set in the specification

![image](https://user-images.githubusercontent.com/5501911/154302814-59a96101-1843-46c4-b343-2b6b590d89e5.png)

**Actual result**

![image](https://user-images.githubusercontent.com/5501911/154298383-3c80ed93-92ee-4d55-ad13-f1c0f4b77984.png)

**Steps to reproduce**

Start from the AsyncAPI studio and add the ` "pattern": "[A-Z]{3}"` in any field to see the rendering

**Troubleshooting**

To solve the issue, simply remove the `lowercase` css class in the https://github.com/asyncapi/asyncapi-react/blob/next/library/src/components/Schema.tsx (line 170)

magicmatatjahu:
Fixed by https://github.com/asyncapi/asyncapi-react/pull/550

PR #550, opened 2022-02-16
fix: remove lowercase in pattern

M3lkior:
Related to #549


magicmatatjahu:
@M3lkior Thanks for PR! It looks good. However, could you push empty commit? I see that you created PR with `master` destination and then switch to the `next` and SonerCloud didn't recognize that and create analize based on `master`.

magicmatatjahu (approved):
LGTM!

PR #595, opened 2022-05-08
feat: show operation summary on sidebar

sarisia:
**Description**

Prior to [@asyncapi/html-template@0.21.0](https://github.com/asyncapi/html-template/releases/tag/v0.21.0),
operations' summary is shown in sidebar but current version does not:

(code from v0.20.1)

https://github.com/asyncapi/html-template/blob/d2c4513eef9b33bfc977ed5ed92eb188e8988b62/partials/sidebar.html#L50-L58

- if operation has a summary, show it on sidebar

**Current behavior**

image

**Updated behavior**

image

**Related issue(s)**

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ A official [React](https://reactjs.org/) component for AsyncAPI 2.0 specificatio
 Run this command to install the component in your project:
 
 ```sh
-npm install --save @asyncapi/react-component
+npm install --save @asyncapi/react-component@next
 ```
 
 Check out this simple sandbox application that uses the React component:
--- a/docs/configuration/config-modification.md
+++ b/docs/configuration/config-modification.md
@@ -20,10 +20,15 @@ interface ConfigInterface {
     schemas?: boolean;
     errors?: boolean;
   };
+  expand?: {
+    messageExamples?: boolean;
+  },
   sidebar?: {
     showOperations?: 'byDefault' | 'bySpecTags' | 'byOperationsTags';
   },
   parserOptions?: ParserOptions;
+  publishLabel?: string;
+  subscribeLabel?: string;
 }
 ```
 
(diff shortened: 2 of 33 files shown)


magicmatatjahu (changes requested):
@sarisia Thanks for PR! Looks great, however, I suggest to display with summary also channel name to identify which channel is described, like `{channelName} {optionalSummary}`.

sarisia:
@magicmatatjahu Thank you for reviewing and sorry for my late response.

I think adding channel name to sidebar makes the line length too long and users may feel hard to find
operations at a grance.

At [Redoc (OpenAPI docs generator)](https://redocly.github.io/redoc/#tag/pet/operation/newPet) they only show summary
on sidebars for years , so I think it's enough to do so.

How do you think?

magicmatatjahu:
@sarisia I would also prefer to add the name of the channel. Here is what it looks like for example. 

image

WDYT?

magicmatatjahu (approved):

magicmatatjahu:
@sarisia Thanks for contribution!

sarisia:
> https://github.com/asyncapi/asyncapi-react/pull/595#issuecomment-1166993545
> @sarisia I would also prefer to add the name of the channel. Here is what it looks like for example. 
> 
> 
> 
> image
> 
> 
> 
> WDYT?

Sorry for not responding for a long time.

Maybe make it controllable by sidebar option is better, but since I'm not the React guy, it may take some time to make PR...

Maybe I'll create an new follow up issue later. Thanks!

magicmatatjahu:
@sarisia No problem! Yeah, it can be :) Please create issue and I will make PR.  I would like to close as many issues as possible in the next few days/weeks, because later I will not have time to do it.

sarisia:
@magicmatatjahu #701, thanks!

shanewazabbas:
I don't like this change. If a summary is really long, it makes the sidebar look like a mess...

sarisia:
> If a summary is really long, it makes the sidebar look like a mess...

Indeed. We had same problem for very long channel name.

#701 suggests adding sidebar option to change the priority between channelName and summary.

PR #604, opened 2022-05-31
feat: add minor improvements

### LINKED ISSUE #440: Ability to auto-expand message schema

### LINKED ISSUE #441: Set the left menu collapsable
#### Reason/Context
Please try answering few of those questions
- Why we need this improvement?

Depending on the screen resolution, the menu on the right side overlay the spec content
- How will this change help?
Allow the user to collapse this menu if needed
- What is the motivation? 
A degraded user experience

image (13)

#### Description
Please try answering few of those questions
- What changes have to be introduced?
A new button to collapse the menu with a new collapsed style (a vertical layout to allow the expand ?)
- Will this be a breaking change?
no
- How could it be implemented/designed?

A vertical layout with a collapse / expand icon

image

### LINKED ISSUE #599: Anchor-link to a specific channel
**Description**

It would be very convenient to be able to create link to specific channel in documentation.

By opening link with channel anchor asyncapi-docs should open appropriate accordion-item and scroll to the channel.

**Reasons**

Convenience.

**Attachments**

eg: https://example.com/asyncapi/#payments-order-paid

### LINKED ISSUE #605: Array of objects has  wrong uid in generated HTML
#### Describe the bug
Array of objects has  uid of object scheme instead of uid of array scheme

#### How to Reproduce
```
{
    "asyncapi": "2.0.0",
    "info": {
        "title": "Bug reproduction",
        "version": "1.0.0",
        "description": "Payload has wrong uid in HTML documentation, when Payload is an array of objects."
    },
    "channels": {
        "bug": {
            "publish": {
                "message": {
                    "$ref": "#/components/messages/Message"
                }
            }
        }
    },
    "components": {
        "messages": {
            "Message": {
                "name": "Message",
                "payload": {
                    "$ref": "#/components/schemas/Payload"
                }
            }
        },
        "schemas": {
            "Payload": {
                "description": "BUG: uid should be Payload but is Object",
                "items": {
                    "$ref": "#/components/schemas/Object"
                }
            },
            "Object": {
                "type": "object"
            }
        }
    }
}
```
image

#### Expected behavior
uid to be set according to schema. In the example above should be "Payload" instead of "Object"
image

magicmatatjahu:
**Description**

- Set the left menu collapsable:
  - expanded:
    
    image
    
  - collapsed
  
    image
  
- add collapse button for nested subschemas in the Schema component:

  image
  image
  
- add operation's summary to the left navigation:
  
  image

- add anchors in headers:

  image
  image

**Related issue(s)**
Resolves #441
Resolves #440
Resolves #599
Resolves #605 

Preview: https://3000-magicmatatj-asyncapirea-pl74cof8gqr.ws-eu46.gitpod.io/
Gitpod url (you need to login by github accout): https://gitpod.io/#https://github.com/asyncapi/asyncapi-react/pull/604

cc @mcturco @derberg

### DIFF
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,2 +1,2 @@
 # Disable cognitive complexity in files
-sonar.exclusions=library/src/components/Schema.tsx
+sonar.exclusions=library/src/components/Schema.tsx,library/src/containers/Sidebar/Sidebar.tsx
--- a/docs/configuration/config-modification.md
+++ b/docs/configuration/config-modification.md
@@ -22,6 +22,7 @@ interface ConfigInterface {
   };
   sidebar?: {
     showOperations?: 'byDefault' | 'bySpecTags' | 'byOperationsTags';
+    collapsed?: boolean;
   },
   parserOptions?: ParserOptions;
 }
@@ -41,6 +42,7 @@ interface ConfigInterface {
 
   This field contains configuration responsible for the way of working of the sidebar.
   `showOperations` field is set to `byDefault` by default.
+  `collapsed` is set to `false` by default.
 
 - **parserOptions?: ParserOptions**
 
@@ -117,6 +119,7 @@ In the above examples, after concatenation with the default configuration, the r
   },
   sidebar: {
     showOperations: 'bySpecTags',
+    collapsed: false,
   },
 }
 ```
--- a/docs/features/anchors.md
+++ b/docs/features/anchors.md
@@ -10,12 +10,12 @@ Each component has a unique anchor in the `http://{URL}/#{SCHEMA_NAME}-{CONTAINE
 
 - `{URL}` is the website URL.
 - `{SCHEMA_NAME}` is the name of the schema. It is passed to the component by configuration with the [`schemaID`](../configuration/config-modification.md#definition) field. By default, the value of `schemaID` is set to an empty string.
-- `{CONTAINER}` is the name of the container. The possible values are: `info`, `servers`, `operations`, `messages`, `schemas`. If `ITEM_NAME` is also present then container has singular form like `operation`, `message` etc.
+- `{CONTAINER}` is the name of the container. The possible values are: `introduction`, `servers`, `operations`, `messages`, `schemas`. If `ITEM_NAME` is also present then container has singular form like `operation`, `message` etc.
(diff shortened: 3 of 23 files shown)


derberg:
> I wouldn't want to split this into 5 separate PRs. I know about problems for reviewers, but the changes are very small, but due to having to do an update of the styles, html as well as the js itself there is much more changed code than there should be. For example, anchors resulted in update of several components because I had to add new functionality for each.

yeah, you can definitely get some inspiration from @jonaslagoni and NATS template https://github.com/asyncapi/dotnet-nats-template/pulls?q=is%3Apr+is%3Aclosed. He did a pro work there, separate PR per each feature and each fix

> I don't want to add special prop for something like this and don't know why it looks bad

I wrote it doesn't look good, not that it looks bad 😆 it just looks cluttered. Left nav is not very "wide" so descriptions make it look super cluttered. This is why I vote for prop that by default disables it. We should not throw something like this at users

fmvilas:
@magicmatatjahu I see you're splitting this PR already, which is awesome 🙏 Please don't forget to close this one.

PR #614, opened 2022-06-27
fix: define react and react-dom as peerDependencies

### LINKED ISSUE #609: Declare react as dependency
As reported by yarn pnp:

```
➜  yarn ag build/asyncapi.yml @asyncapi/html-template -o ./docs --force-write
Something went wrong:
Error: @asyncapi/react-component tried to access react, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.
```

magicmatatjahu:
**Description**

We use `react` and `react-dom` in component, but we don't have defined these packages as peer-dependencies and e.g. yarn throw error due to this fact. See more #609 

I tested package installing new project with yarn without and with `react/react-dom` deps and yarn in first case throw errors, in second installation passed.

**Related issue(s)**
Resolves #609

### DIFF
--- a/library/package.json
+++ b/library/package.json
@@ -77,6 +77,10 @@
     "openapi-sampler": "^1.2.1",
     "use-resize-observer": "^8.0.0"
   },
+  "peerDependencies": {
+    "react": ">=16.8.0",
+    "react-dom": ">=16.8.0"
+  },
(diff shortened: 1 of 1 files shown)


magicmatatjahu:
@derberg ping, ping :)

Issue #649, opened 2022-09-29
Group servers in sidebar navigation like Operation Objects


magicmatatjahu:
**Description**

As in title. I think it's clear :) We added support for `tags` in Server Objects in 2.5.0 version of spec.

PR #682, opened 2022-12-21
feat: group Servers in sidebar navigation like Operation Objects

aeworxet:
This PR adds grouping of servers by tags in sidebar navigation like Operation Objects.

Grouping of servers by tags is made on helper's level, so that `serversTags` receives map, in which servers are already grouped by tags
```
const serversTags = SpecificationHelpers.serversTags(asyncapi);
```
(more complex structure than array, with more levels of nesting is chosen, because each server can have multiple tags which can also overlap)

image

Further, `ServersByTags()` gets tag, groups all servers that have it under the name of the tag, repeats.
Then it groups untagged servers.

There was no real benefit in writing untagged servers' grouping using intermediate array. It is written this way only to add syntactic sugar, by aligning with the way grouping of tagged servers is written.

Fixes https://github.com/asyncapi/asyncapi-react/issues/649

### DIFF
--- a/library/src/helpers/specification.ts
+++ b/library/src/helpers/specification.ts
@@ -50,15 +50,19 @@ export class SpecificationHelpers {
   /**
    * Check if given schema have one of the specified tags.
    */
-  static containTags(schema: any, tags: Tag | Tag[]): boolean {
+  static containTags(
+    schema: { tags: () => Tag[] },
+    tags: string | string[],
+  ): boolean {
     const tagsToCheck =
       typeof schema.tags === 'function' ? schema.tags() : undefined;
-    if (tagsToCheck === undefined || !Array.isArray(tagsToCheck)) {
+    if (!tagsToCheck || !Array.isArray(tagsToCheck)) {
       return false;
     }
-    tags = Array.isArray(tags) ? tags : [tags];
+
+    const arrayTags = Array.isArray(tags) ? tags : [tags];
     return tagsToCheck.some((tag: Tag) =>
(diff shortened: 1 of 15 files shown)


library/src/helpers/specification.ts
```diff
@@ -79,4 +79,23 @@ export class SpecificationHelpers {
...
+          if (tags[tag.name()]) {
+            tags[tag.name()] = [tags[tag.name()], _];
+          } else {
+            tags[tag.name()] = _;
+          }
+        });
+      }
+    });
+    return tags;
+  }
```
magicmatatjahu, line 100: You try to make map of tags, but you should make rather array of tags, so please similar function to the `operationsTags`:

```suggestion
  static serversTags(spec: AsyncAPIDocument) {
    const tags = new Map<string, Tag>();
    Object.entries(spec.servers()).forEach(([_, server]) => {
      if (server.hasTags()) {
        server.tags().forEach(tag => tags.set(tag.name(), tag));
      }
    });
    return Array.from(tags.values());
  }
```

magicmatatjahu (changes requested):
Good :) I added 3 suggestions.

derberg (commented):
@aeworxet even though it was not you that added duplicated code, sonar picked that up as you added some modifications to duplication-related-lines. This is why it is pinging you for changes.

IMHO it is ok if you report a separate issue for it and leave it out of scope for this PR

magicmatatjahu (approved):

magicmatatjahu:
@aeworxet Thanks for contribution!

PR #715, opened 2023-04-12
feat: add support for mermaid diagrams in markdown

### LINKED ISSUE #712: Add mermaid diagram support to markdown
**Description**

It would be great to include mermaid support (flow charts, sequence diagrams, class diagrams, etc.) in markdown.

**Reasons**

Diagrams are useful in documentation.

**Initial Research**

It looks like `asyncapi-react` already uses `markdown-it` and [this VS code extension](https://github.com/mjbvz/vscode-markdown-mermaid) demonstrates how easy it can be to add mermaid support to `markdown-it` containers.

j-h-a:
**Description**

This change adds support to include mermaid diagrams directly in any markdown.

**Related issue(s)**

Fixes #712

### DIFF
--- a/library/src/helpers/marked.ts
+++ b/library/src/helpers/marked.ts
@@ -15,7 +15,7 @@ hljs.registerLanguage('yaml', yaml);
 import bash from 'highlight.js/lib/languages/bash';
 hljs.registerLanguage('bash', bash);
 
-const markedOptions: marked.MarkedOptions = {
+const codeHighlightOptions: marked.MarkedOptions = {
   langPrefix: 'hljs language-',
   highlight: (code, language) => {
     if (!hljs.getLanguage(language)) {
@@ -28,9 +28,30 @@ const markedOptions: marked.MarkedOptions = {
     }
   },
 };
+const codeHighlightRenderer = new marked.Renderer(codeHighlightOptions);
+const markdownRenderer = new marked.Renderer();
+markdownRenderer.code = (code, language, isEscaped) => {
+  if (language === 'mermaid') {
+    addMermaidIfNeeded();
+    return '<pre class="mermaid">' + code + '</pre>';
+  }
+  return codeHighlightRenderer.code(code, language, isEscaped);
+};
+
+let gotMermaidAlready = false;
+function addMermaidIfNeeded() {
+  if (gotMermaidAlready) {
+    return;
+  }
+  gotMermaidAlready = true;
+  const script = document.createElement('script');
+  script.type = 'module';
+  script.text = `import mermaid from 'https://cdn.jsdelivr.net/npm/mermaid@10/dist/mermaid.esm.min.mjs';`;
+  document.body.appendChild(script);
+}
 
 export function renderMarkdown(content: string): string {
-  return marked(content, markedOptions);
+  return marked(content, { renderer: markdownRenderer });
 }
 
 export { hljs };
--- a/library/src/helpers/marked.ts
+++ b/library/src/helpers/marked.ts
@@ -1,4 +1,5 @@
 import { marked } from 'marked';
+import { initializeMermaidOnce }  from './mermaid';
 
 // @ts-ignore
 import hljs from 'highlight.js/lib/core';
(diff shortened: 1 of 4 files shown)


derberg:
@j-h-a thanks for that PR 🍺 and sorry again for huge delay 😞 

Just to clarify, we are not using `makrdown-it` in react component but `marked`. Just keep in mind for future that in the main library we can only use libraries that are explicitly defined as dependencies -> https://github.com/asyncapi/asyncapi-react/blob/next/library/package.json#L70-L79. I think you just looked into `master` first.

It is the same with `mermaid`. I assume you've hacked it this way with global variable and dynamic script addition to not add to much more KB to bundled JS files. The problem is that this way we do not control `mermaid` version through package.json and will not get any potential automated dependabot updates of patch security fixes. 
- please just install `mermaid` package in `library` folder
- just import it as any other extenral library inside `marked.ts`
- adjust your code a little bit to match new setup
- instead of making sure the script is there already, rather make sure mermaid was already initialized once and do not need to be initialized.
- you have some code for inspiration in https://github.com/asyncapi/website/blob/master/components/MDX.js

After refactor you can check the size of bundled JS file and I bet it did not increase too much.

---

Regarding tests, we have simple tests and in case of your change just add a test case to https://github.com/asyncapi/asyncapi-react/blob/next/library/src/helpers/__tests__/marked.test.ts called `should render mermaid diagram`.

---

Some magic in tailwind configuration has to probably be introduced as by default the background of the diagram should be white. I think now default background for `pre` tags is applied

---

Also would make sense to mention somewhere in the readme support for mermaid

library/src/helpers/marked.ts
```diff
@@ -28,9 +29,18 @@ const markedOptions: marked.MarkedOptions = {
     }
   },
 };
+const codeHighlightRenderer = new marked.Renderer(codeHighlightOptions);
+const markdownRenderer = new marked.Renderer();
+markdownRenderer.code = (code, language, isEscaped) => {
+  if (language === 'mermaid') {
+    initializeMermaidOnce();
+    return '<pre class="mermaid">' + code + '</pre>';
```
jonaslagoni, line 37: Dont you need to return the rendered mermaid diagram here? i.e. [add this from the website](https://github.com/asyncapi/website/blob/master/components/MDX.js#L119):
```ts
let currentId = 0;
const uuid = () => `mermaid-${(currentId++).toString()}`;
function MermaidDiagram({ graph }) {
  const [svg, setSvg] = useState(null);

  useLayoutEffect(() => {
    if (!graph) {
      return;
    }

    try {
      mermaid.mermaidAPI.render(uuid(), graph, renderedSvg => setSvg(renderedSvg));
    } catch (e) {
      setHtml(null)
      console.error(e);
    }
  }, [graph]);

  return svg ? <div dangerouslySetInnerHTML={{ __html: svg }} /> : null;
}
```

And then do the following:
```suggestion
    return <MermaidDiagram graph={code} />;
```

library/src/helpers/marked.ts
```diff
@@ -28,9 +29,18 @@ const markedOptions: marked.MarkedOptions = {
     }
   },
 };
+const codeHighlightRenderer = new marked.Renderer(codeHighlightOptions);
+const markdownRenderer = new marked.Renderer();
+markdownRenderer.code = (code, language, isEscaped) => {
+  if (language === 'mermaid') {
+    initializeMermaidOnce();
```
jonaslagoni, line 36: Remove this.

library/src/helpers/marked.ts
```diff
@@ -15,7 +16,7 @@ hljs.registerLanguage('yaml', yaml);
 import bash from 'highlight.js/lib/languages/bash';
 hljs.registerLanguage('bash', bash);
```
jonaslagoni, line 18: Globally inline everything, just as the website does it:

```suggestion
let mermaidInitialized = false;
initializeMermaid();
function initializeMermaid() {
  if (mermaidInitialized) {
    return;
  }

  mermaidInitialized = true;
  mermaid.initialize({
    startOnLoad: true,
    theme: "base",
    securityLevel: "loose",
    themeCSS: ``,
    themeVariables: {
      primaryColor: "#EDFAFF",
      primaryBorderColor: "#47BCEE",
      secondaryColor: "#F4EFFC",
      secondaryBorderColor: "#875AE2",
      fontFamily: "Inter, sans-serif",
      fontSize: "18px",
      primaryTextColor: "#242929",
      tertiaryColor: "#F7F9FA",
      tertiaryBorderColor: "#BFC6C7"
    }
  });
}
```

This will most likely remove the error you have, about re-exporting the initializeMermaidOnce.

jonaslagoni (commented):
Just left some suggestions 🙂 They are rough code suggestion, you might need to alter them slightly and add some dependencies.

derberg:
`next` will soon be deleted. It is already merged into `master` and 1.0 is now available.

 no more long living release branches!

sorry for any trouble with your PR 😞 but we really had to finally release this v1 and get rid of release branch

I close this PR as I think it is better for you to recreate a new fresh PR targeting `master`. Feel free to drop a comment if you disagree and I will reopen ☮️

PR #742, opened 2023-07-28
chore: Set Access-Control-Allow-Origin header for playground devserver

Laupetin:
**Description**

- Set "Access-Control-Allow-Origin" header for playground dev server to "*" to allow playground to fetch an external schema from somewhere other than same origin when running it locally.

image
(Trying to load schema from `http://localhost:8090/docs/asyncapi.json` with playground devserver running on `http://localhost:3000` for example)

**Related issue(s)**

### DIFF
--- a/playground/config-overrides.js
+++ b/playground/config-overrides.js
@@ -1,9 +1,19 @@
 const path = require('path');
 
-module.exports = function override(config, env) {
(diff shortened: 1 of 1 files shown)


Laupetin:
May have misunderstood what i was doing wrong, my bad

PR #746, opened 2023-08-25
feat!: update to new parser api

jonaslagoni:
**Description**
This PR updates the library to the newest parser API, and effectively adds support for v3 (at least it can understand v3).

### DIFF
--- a/library/src/components/Schema.tsx
+++ b/library/src/components/Schema.tsx
@@ -1,12 +1,12 @@
 import React, { useState, useEffect, useContext } from 'react';
-import { Schema as SchemaType } from '@asyncapi/parser';
+import { SchemaInterface } from '@asyncapi/parser';
 
 import { Href, CollapseButton, Markdown, Extensions } from './index';
 import { SchemaHelpers } from '../helpers';
 
 interface Props {
   schemaName?: React.ReactNode;
-  schema?: SchemaType;
+  schema?: SchemaInterface;
   required?: boolean;
   isPatternProperty?: boolean;
   isProperty?: boolean;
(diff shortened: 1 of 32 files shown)


library/src/components/Schema.tsx
```diff
@@ -1,12 +1,12 @@
 import React, { useState, useEffect, useContext } from 'react';
```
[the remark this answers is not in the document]
magicmatatjahu: `schema.isCircular()`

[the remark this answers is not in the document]
jonaslagoni: I am using that, but it does not interact well with the current setup in the file.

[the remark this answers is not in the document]
magicmatatjahu: Which setup? Do you mean cases with `x-parser-circular-props` or similar? It's not needed anymore. Remove all tests which has logic based on that and leave only with checking by `isCircular()`

library/src/helpers/parser.ts
```diff
@@ -1,27 +1,26 @@
...
 
 import { VALIDATION_ERRORS_TYPE } from '../constants';
 
-registerSchemaParser(openapiSchemaParser);
-registerSchemaParser(avroSchemaParser);
-registerSchemaParser(protoSchemaParser);
+const asyncapiParser = new AsyncapiParser();
+asyncapiParser.registerSchemaParser(OpenAPISchemaParser());
+asyncapiParser.registerSchemaParser(AvroSchemaParser());
+//asyncapiParser.registerSchemaParser(protoSchemaParser);
```
jonaslagoni, line 14: Until proto parser is updated to newest parser API it will be disabled here.

library/src/components/__tests__/Schema.test.tsx
```diff
@@ -24,34 +22,8 @@ describe('Schema component', () => {
...
-        },
-        circular: {},
-        circularTwo: {},
-      },
-    };
-    schema.properties.circular = schema;
-    schema.properties.circularTwo = schema;
-    schema['x-parser-circular-props'] = ['circular', 'circularTwo'];
-    const schemaModel = new SchemaModel(schema);
+    schema['x-schema-private-render-type'] = true;
```
jonaslagoni, line 25: I had to add this to render the type for some reason, even though it was not needed in the previous setup. 

I simply have no idea where this is supposed to already have been sat, any idea @magicmatatjahu?

library/src/components/Schema.tsx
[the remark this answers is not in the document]
jonaslagoni: Yea thats definitely easier 😄 

Have one leftover "problem": https://github.com/asyncapi/asyncapi-react/pull/746/files#r1305679088

library/src/components/__tests__/Schema.test.tsx
magicmatatjahu: I don't remember exactly, but it shouldn't be there, maybe this `if` (https://github.com/asyncapi/asyncapi-react/blob/8cf921d8c1e5b231222d615d242f62144146049e/library/src/components/Schema.tsx#L47) is needed for some extra stuff, I seriously don't remember

library/src/helpers/parser.ts
fmvilas: Would be cool to leave that comment in the code.

jonaslagoni: Should have it's own issue after this is merged 🙂

library/src/components/__tests__/Schema.test.tsx
jonaslagoni: So I checked the old code, parser v1.8, and new code and found 0 references to `x-schema-private-render-type` and it being set to `true`.

So, I removed the entire extension in https://github.com/asyncapi/asyncapi-react/pull/746/commits/770197296de88118bf80de637025a8ea372716f8

library/src/helpers/parser.ts
jonaslagoni: Added a comment as well

jonaslagoni:
FYI, I had to update to webpack 5 to make it work, but don't see that as a downside 😆

jonaslagoni:
@fmvilas @derberg @magicmatatjahu all tests finally pass and it builds correctly, please have a look when you can 🙏

fmvilas (approved):
LGTM 👍

fmvilas:
Are you afraid of something? 😄

PR #749, opened 2023-08-31
feat: add request and reply view

jonaslagoni:
**Description**
This PR adds request and reply information to be rendered.


jonaslagoni:
Recreated in https://github.com/asyncapi/asyncapi-react/pull/756

PR #758, opened 2023-09-06
Feat: re add protobuff parser

GreenRover:
re enable protobuff parser lib

PR #759, opened 2023-09-07
feat: support protobuf

GreenRover:
This is a spinn of from #758

Because in this project there was package-lock.js in format 1 and 2 i have had trouble to find the nodejs version to create only minimal changes. Hop this works for you.


GreenRover:
anyone an idea why test are failing?

derberg:
`next` will soon be deleted. It is already merged into `master` and 1.0 is now available.

 no more long living release branches!

I don't think we need this PR anymore

PR #766, opened 2023-09-12
fix: sidebar problems

jonaslagoni:
**Description**
The sidebar changes were not being tested anywhere, and hence the problems were not caught until I tried using them in https://github.com/asyncapi/html-template/pull/448.

This PR fixes those problems and adds a test to ensure that all shown options work.


fmvilas (approved):

fmvilas:
Good catch mate!

PR #773, opened 2023-09-27
ci: eject from global release config file

derberg:
Release of 1.0 failed -> https://github.com/asyncapi/asyncapi-react/actions/runs/6322951494

We can use global generic release workflow, but not the config, cause of monorepo. 

`get-global-releaserc` topic already removed


fmvilas (approved):

derberg:
jizzzz forgot it will not trigger a release 😭

PR #775, opened 2023-09-28
fix: playground dependencies need to use `file` to fix installation

derberg:
release went well but version bump failed

made an inline comment that fixes it


magicmatatjahu (approved):
LGTM!

PR #795, opened 2023-10-02
feat: add support for request and reply pattern

AceTheCreator:
This PR adds support for request and reply patterns by introducing a new component that renders reply information such as:
- reply channel information such as `address`, `description`, `message(s)` and so on
- Reply address information such as `address`, `location`, and `description`

https://github.com/asyncapi/asyncapi-react/assets/40604284/3228e97f-3132-4569-8b23-765cacf9bc28

This PR also fixes the PUB/SUB reversal experienced in this issue #815 

cc @jonaslagoni @magicmatatjahu


library/src/containers/Messages/Message.tsx
```diff
@@ -49,6 +49,16 @@ export const Message: React.FunctionComponent<Props> = ({
   const externalDocs = message.externalDocs();
   const showInfoList = contentType || externalDocs;
 
+  // For some reason the parser is returning the wrong message ID,
+  // when retrieving reply() messages that's the reason retrieving it this way
```
derberg, line 53: smells like a parser bug that you should not workaround but check in parser

derberg (changes requested):
can you share an example that you tested it against, and share screen shots as result of using official 3.0 adeo and kraken examples? It is strange for me, the part that `reply should be done on channel` especially that sometimes it is dynamic info, and you do not really know what channel it is

AceTheCreator:
> the part that `reply should be done on channel` especially that sometimes it is dynamic info, and you do not really know what channel it is

Yea, the result is also dynamic 🤔

library/src/containers/Messages/Message.tsx
AceTheCreator: Okay, I'll create an issue in the parser

AceTheCreator:
@derberg Here's what it's like testing it with Kraken example

https://github.com/asyncapi/asyncapi-react/assets/40604284/133c3a19-4366-4476-8ae8-a2f6e217b5a1

library/src/containers/Messages/Message.tsx
fmvilas: Did you create the issue in the parser? I'm trying to find it without luck.

fmvilas (approved):
Looks good to me in general. Would you mind opening the issue in the parser? That's all I'd ask.

AceTheCreator:
@derberg @fmvilas @jonaslagoni  I removed the way I was retrieving the message ID, and below is what the UI is like using adeo example(I also added a dummy description to the reply address object)

https://github.com/asyncapi/asyncapi-react/assets/40604284/384219ac-5466-4c06-8634-3f92ec8d4f0a

The one below includes a message

https://github.com/asyncapi/asyncapi-react/assets/40604284/9b81a405-34d8-4ae9-9583-135469e27c91

fmvilas (approved):

fmvilas (approved):
Looks good to me @AceTheCreator. We can polish the design and the wording but we have a pending task to do so for the whole document, so don't worry much about it. Great stuff! 🚀

derberg (changes requested):
tried on local with playgroud this example: https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/kraken-websocket-request-reply-message-filter-in-reply-asyncapi.yml

left some notes 👇🏼 
Screenshot 2023-10-30 at 14 10 16

I also think that if the operation has `title` we should show below not the address of the operation, but the human readable title 🤔 

Screenshot 2023-10-30 at 14 13 36

and last but not least, SUB and PUB should become RECEIVES and SENDS - only for documents 3.x

@AceTheCreator @fmvilas you can already have a look, I will check with other examples but in the meantime @AceTheCreator can work on fixes that I pointed

derberg:
I can also confirm this PR fixes https://github.com/asyncapi/asyncapi-react/issues/815 so @AceTheCreator please update PR description to make sure this issue will be closed after merge

derberg:
ah, and also from parser perspective, shouldn't we use [3.0.0-next-major-spec.7](https://github.com/asyncapi/parser-js/compare/v3.0.0-next-major-spec.6...v3.0.0-next-major-spec.7) here?

AceTheCreator:
@derberg, regarding your question about why the parser is returning an array index, I created an issue in the parser I'd suggest you take a look https://github.com/asyncapi/parser-js/issues/873

About replacing the address with the title if it exists, do we really have to do that? and if yes, how should it be presented in the UI

All suggested changes have been added and I also need help looking into why sonarcloud is failing for some weird reason 🤔

cc @derberg @fmvilas @jonaslagoni

derberg:
> About replacing the address with the title if it exists, do we really have to do that? and if yes, how should it be presented in the UI

can be a followup later.

> @derberg, regarding your question about why the parser is returning an array index, I created an issue in the parser I'd suggest you take a look https://github.com/asyncapi/parser-js/issues/873

thanks!

regarding code smells, in case of array index `<li className="mt-4" key={idx}>` you could use msg.id

---
heading for final review

derberg:
@AceTheCreator how do you test changes on your side? I pulled your latest branch and did:

```
npm run clean
npm i
npm start
```

and in playground I do not see any changes, still `SUB` in v3 docs, also only `SUB` in prior v3 docs, even if operation is `publish`. Also broken reply panel 

Screenshot 2023-11-06 at 14 18 50

some weird cache or? but in private mode still the same 🤔

AceTheCreator:
> @AceTheCreator how do you test changes on your side? I pulled your latest branch and did:
> 
> ```
> npm run clean
> npm i
> npm start
> ```
> 
> and in playground I do not see any changes, still `SUB` in v3 docs, also only `SUB` in prior v3 docs, even if operation is `publish`. Also broken reply panel
> 
> Screenshot 2023-11-06 at 14 18 50
> 
> some weird cache or? but in private mode still the same 🤔

That's weird, I test my changes using the following approach
```
npm run build
cd playground
npm run start
```

derberg:
dunno but can't refresh locally 🤷🏼 
luckily I made it with gitpod: https://gitpod.io/#https://github.com/AceTheCreator/asyncapi-react/tree/req/rep

---

tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/kraken-websocket-request-reply-multiple-channels-asyncapi.yml

issue: reply for `subscribe` operation do not have much channel details for reply, only reply address - missing info about messages. Basically `Expected Reply Messages` is missing

Screenshot 2023-11-06 at 15 42 18

---

tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/kraken-websocket-request-reply-message-filter-in-reply-asyncapi.yml and works like a charm

---

tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/adeo-kafka-request-reply-asyncapi.yml 

issue: Playground crash. Cannot figure errors, strange as latest parser is used and the bug with kafka binding was supposed to be fixed there. Will investigate

---

tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/streetlights-kafka-asyncapi.yml

issue: no operations rendered, just `Operations` header

---

last but not least - I think https://github.com/asyncapi/asyncapi-react/pull/795#issuecomment-1785188840 is broken again

AceTheCreator:
@derberg I added suggested changes. Kindly give it a look :)

derberg:
> tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/adeo-kafka-request-reply-asyncapi.yml
issue: Playground crash. Cannot figure errors, strange as latest parser is used and the bug with kafka binding was supposed to be fixed there. Will investigate

did you have a chance to check what is wrong here?

> tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/streetlights-kafka-asyncapi.yml
issue: no operations rendered, just Operations header

Operations are still empty

---
also some issue with rendering of examples in reply

Screenshot 2023-11-13 at 10 31 33

AceTheCreator:
> tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/streetlights-kafka-asyncapi.yml
issue: no operations rendered, just Operations header

> Operations are still empty

@derberg Based on my investigation, The issue is actually related to the way Parser is handling the Operation traits. I get the list of Operations only when I get rid of the traits.

cc @smoya

AceTheCreator:
> tested with https://raw.githubusercontent.com/asyncapi/spec/next-major-spec/examples/adeo-kafka-request-reply-asyncapi.yml
> issue: Playground crash. Cannot figure errors, strange as latest parser is used and the bug with kafka binding was supposed to be fixed there. Will investigate

> did you have a chance to check what is wrong here?

Same as last time. Playground only renders when I remove the Message level bindings 🤷🏾‍♂️

derberg:
@AceTheCreator I see for message issue you have created https://github.com/asyncapi/parser-js/issues/877, and what about operations? maybe update `message`-related issue or create a new one for `operations`

derberg:
so in this PR only styling of examples left?

library/package.json
```diff
@@ -70,8 +70,8 @@
   "dependencies": {
     "@asyncapi/avro-schema-parser": "3.0.3",
     "@asyncapi/openapi-schema-parser": "3.0.4",
-    "@asyncapi/parser": "^2.1.1",
     "@asyncapi/protobuf-schema-parser": "3.0.0",
+    "@asyncapi/parser": "^3.0.0-next-major-spec.7",
```
derberg, line 74: please use latest [v3.0.0-next-major-spec.10](https://github.com/asyncapi/parser-js/releases/tag/v3.0.0-next-major-spec.10)

playground/package.json
```diff
@@ -28,7 +28,7 @@
     "react-scripts": "3.1.2"
   },
   "scripts": {
-    "start": "react-app-rewired start",
+    "start": "react-app-rewired --openssl-legacy-provider start",
```
derberg, line 31: please revert, as with this I cannot start local playground anymore, and need to manually remove that flag

derberg (changes requested):
2 small comments + please update description to reflect current state of the PR

then we can merge

AceTheCreator:
added suggested changes @derberg

derberg (approved):
awesome hard work!

Issue #785, opened 2023-09-30
Remove this redundant jump.


AceTheCreator:
## Why this issue?
https://sonarcloud.io/project/issues?resolved=false&id=asyncapi_asyncapi-react&open=AYrVnLzdi9mf_H5ePyyQ&tab=why

## Where's the issue? 
https://sonarcloud.io/project/issues?resolved=false&id=asyncapi_asyncapi-react&open=AYrVnLzdi9mf_H5ePyyQ&tab=code

PR #807, opened 2023-10-04
refactor: remove the redundant-jump code

parasss19:
**Description**

-Removed the redundant jump code

**Related issue(s)**
#785

### DIFF
--- a/library/src/helpers/message.ts
+++ b/library/src/helpers/message.ts
@@ -56,8 +56,6 @@ export class MessageHelpers {
     if (payload && payload.examples()) {
       return payload.examples()?.map(example => ({ example }));
     }
-
-    return;
   }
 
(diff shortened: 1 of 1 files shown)


derberg (commented):
signature of functions that you modify say they return `MessageExample[] | undefined` so if you remove `return` tests fail. So please bring back `return` but as `return undefined`

library/src/helpers/message.ts
```diff
@@ -56,8 +56,6 @@ export class MessageHelpers {
     if (payload && payload.examples()) {
       return payload.examples()?.map(example => ({ example }));
     }
-
-    return;
```
derberg, line 60: btw, it was not in scope of this issue, just the other return

derberg (approved):

Issue #822, opened 2023-10-25
Anchors for operations are not working as expected


kennethaasan:
**Description**

**Expected result**

As a user, I want to be able to click and link to operations anchors

**Actual result**

You're not moved to the operation in the URL or when you click on a operation in the sidebar

**Steps to reproduce**

E.g. https://asyncapi.github.io/asyncapi-react/#operation-subscribe-smartylighting/streetlights/1/0/action/{streetlightId}/dim doesn't move you to that operation when you open the link. And if you press the operation from the sidebar, it doesn't work as expected either.

**Troubleshooting**

PR #859, opened 2023-12-12
fix: corrected anchor tags href in Playground sidebar

### LINKED ISSUE #830: HTML anchor broken on sidebar
**Description**

In sidebar, HTML anchor link to operation seems link to unknown id.

**Expected result**

After clicking HTML anchor link on side bar, browser should to open related operation.

**Actual result**

After clicking HTML anchor link on side bar, URL change and nothing happened.

**Steps to reproduce**

- open <https://asyncapi.github.io/asyncapi-react/>
- fetch url from <https://www.asyncapi.com/resources/casestudies/adeo/asyncapi.yaml>
- Click link `SUB
[COSTING] Request one or more Costing calculation for any product`

**Troubleshooting**

Santhoshmani1:
**Description**

- Fixed  anchor links in playground Sidebar by adding id to the operations.

**Related issue(s)**

**Fixes**
-  #830 HTML anchor broken on sidebar
-  #822 Anchors for operations are not working as expected

### DIFF
--- a/library/src/containers/Sidebar/Sidebar.tsx
+++ b/library/src/containers/Sidebar/Sidebar.tsx
@@ -360,7 +360,7 @@ const OperationsList: React.FunctionComponent = () => {
 interface OperationItemProps {
   channelName: string;
   summary: string;
-  kind: 'publish' | 'subscribe';
+  kind: 'subscribe' | 'publish';
 }
 
 const OperationItem: React.FunctionComponent<OperationItemProps> = ({
@@ -383,7 +383,7 @@ const OperationItem: React.FunctionComponent<OperationItemProps> = ({
     <li>
       <a
         className="flex no-underline text-gray-700 mb-2 hover:text-gray-900"
-        href={`#operation-${kind}-${channelName}`}
+        href={'#' + channelName}
         onClick={() => setShowSidebar(false)}
       >
         <span
--- a/library/src/containers/Operations/Operation.tsx
+++ b/library/src/containers/Operations/Operation.tsx
@@ -241,6 +241,7 @@ export const OperationInfo: React.FunctionComponent<Props> = props => {
           <span
             className={`font-mono border uppercase p-1 rounded mr-2 ${borderColor}`}
             title={type}
+            id={channelName}
           >
             {typeLabel}
           </span>{' '}


BTOdell:
This doesn't appear to fix the problem correctly. You're adding a new anchor instead of fixing the existing anchor with the format: `#operation-{label}-{channel}`.

derberg:
@Santhoshmani1 please have a look at comment from @BTOdell

Santhoshmani1:
https://github.com/asyncapi/asyncapi-react/assets/119673958/38569d09-61c7-4042-8e3a-fc7c76cb9318

I have changed the href attribute of sidebar links and added the id attribute to match in the operations. It's working correctly as expected and fixes the problem.

library/src/containers/Sidebar/Sidebar.tsx
```diff
@@ -360,7 +360,7 @@ const OperationsList: React.FunctionComponent = () => {
 interface OperationItemProps {
   channelName: string;
   summary: string;
-  kind: 'publish' | 'subscribe';
+  kind: 'subscribe' | 'publish';
```
derberg, line 363: why is this change needed?

library/src/containers/Sidebar/Sidebar.tsx
```diff
@@ -383,7 +383,7 @@ const OperationItem: React.FunctionComponent<OperationItemProps> = ({
     <li>
       <a
         className="flex no-underline text-gray-700 mb-2 hover:text-gray-900"
-        href={`#operation-${kind}-${channelName}`}
+        href={'#' + channelName}
```
derberg, line 386: @AceTheCreator do you recall how will it affect v2 vs v3 compatibility?

@Santhoshmani1 I see in your video that you text with AsyncAPI v2. Can you see how it works with v3 -> https://github.com/asyncapi/spec/blob/master/examples/streetlights-kafka-asyncapi.yml

I'm not sure change is so trivial as in case of v2 operations are part of channels, while in v3 they are not, and there is no pub/sub kinds 🤔 

nevertheless for sure it is not enough just to fix href here. For example when you will check the v3 document that I shared with you, the operation id is `operation-receive-receiveLightMeasurement` where `receive` is `kind`, and `receiveLightMeasurement` is not `channelName` but it is the name of operation. You probably need to extend `OperationItem` component to also have access to operation name, and us it in href 🤔

derberg (changes requested):

kennethaasan:
@Santhoshmani1, I just wanted to say thank you for the contribution 🚀

PR #889, opened 2024-01-18
fix: remove unnecessary playground scripts and update package.json

### LINKED ISSUE #888: Workflows failing because of some leftovers scripts
**Version bump workflow do not work anymore**

This PR https://github.com/asyncapi/asyncapi-react/pull/887 introduced some different way of handling dependency info to react component inside playground

`@asyncapi/react-component": "file:../library`

Since playground is not released as separate package, and people instead should use `asyncapi/studio` - this change can stay, we just need some scripts cleanup:

TODO:
- script https://github.com/asyncapi/asyncapi-react/blob/master/playground/package.json#L34 must be removed
- script file https://github.com/asyncapi/asyncapi-react/blob/master/playground/bump-react-comp.sh must be removed
- modify https://github.com/asyncapi/asyncapi-react/blob/master/package.json#L34 to not trigger removed playground install script

vishvamsinh28:
removed unneccessary scripts

fixes #888


package.json
```diff
@@ -31,7 +31,6 @@
     "bump:playground:version": "cd playground && npm --no-git-tag-version --allow-same-version version $VERSION",
     "bump:webcomp:version": "cd web-component && npm --no-git-tag-version --allow-same-version version $VERSION",
     "bump:lib:version": "cd library && npm --no-git-tag-version --allow-same-version version $VERSION",
-    "bump:version": "npm run bump:lib:version && npm run bump:webcomp:version && npm run bump:playground:version && npm run install:reactcomp:webcomponent && npm run install:reactcomp:playground",
     "install:reactcomp:playground": "cd playground && npm run install:reactcomp",
```
derberg, line 34: oh, this must be removed to

vishvamsinh28: Line 35 ?

vishvamsinh28:
@derberg updated the script

jonaslagoni:
ping ping @derberg

PR #744, opened 2023-08-06
feat: use schema's title when available in `{one,any,all}Of` applicators

KtorZ:
**Description**

Applicators can be a little hard to navigate, especially on deeply nested schema. While the `uid` provides a nice middle-ground solution, it doesn't actually work when schemas are defined outside of the main specification and more importantly, it isn't necessarily meant to be human-readable. The `title` property however is meant just for that.

Changes proposed in this pull request:

- This PR does just that, it adds the `title` next to branches enumeration in `{one,any,all}Of` applicators. When not available, it defaults to the current behavior.

image

**Related issue(s)**

N/A

PR #890, opened 2024-01-19
fix: only render suffix in applicatorSchemaName if it exists

kennethaasan:
**Description**

- #744 introduced a bug when schemas don't include a title. It now renders "Consists of undefined:" and "And of undefined:". This fix makes sure the suffix is dropped when the title doesn't exist.

**Related issue(s)**

https://github.com/asyncapi/asyncapi-react/pull/744/files#r1433826404


kennethaasan:
Damn, I made 877. Closing

PR #929, opened 2024-02-29
feat: creates new playground with next.js

kennethaasan:
**Description**

Changes proposed in this pull request:

- creates new playground with next.js
- upgrades typescript

**Related issue(s)**

Partly fixes #891

PR #949, opened 2024-03-14
fix: fixes playground when deployed to github pages

kennethaasan:
**Description**

- fixes playground when deployed to github pages

**Related issue(s)**

Fixes #929


derberg (dismissed):

derberg:
you need to add playground to sonar ignore, we do not need it there to bother us in playground

kennethaasan:
@derberg, I would like to keep SQ if possible, so I changed the replace a bit to make SQ happy. Please review again

derberg (approved):

Issue #819, opened 2023-10-18
React Components fot extensions Catalog


sambhavgupta0705:
**Description**
As part of the "MVP integration of extensions catalog with AsyncAPI tools to make extension catalog useful" mentorhsip program, detailed here https://github.com/asyncapi/extensions-catalog/issues/78. We would like to add react component to visualize the extensions catalog.

**Reasons**

Extensions are part of the spec. So we need to create react component to compile theses extensions as a part of the AsyncAPI spec. So basically, this functionnality allow us to have suggestions on the existed extensions and to be able to add them when we are writing our code.

We would like to know how and where in the code would be the best way to provide extensions-related components.
Now we are working on the twitter extension, we want to expose at the first time this one, you can check it here : https://github.com/asyncapi/extensions-catalog/pull/112 ( we are still working on their format).

Min2who:
Hii @sambhavgupta0705 may i know the progress of the issue . I would love to look into it

magicmatatjahu:
The best place to write components will be there https://github.com/asyncapi/asyncapi-react/blob/master/library/src/components/Extensions.tsx However props to the component (root component, not mentioned in link) should have also option to pass custom components to render another extensions, written by user of component.

My idea:
- component has registry of extension components, like `TwitterExtension` - we can add `extensions` property on the config to the `AsyncAPIComponent`. It should be simple JS object.
- inside `Extensions` component we try to retrieve the proper component for given extension:

  ```ts
  const { extensions } = useConfig()
  const ExtensionComponent = extensions[extension]
  ```
  
  of course we need to change the logic in the current `Extensions` component
  
- pass to the custom component, the extension value, asyncapi document, maybe also config passed to the root component and parent model (where extensions are defined) from parser-js
- when registry won't have custom component, we should render extension as normal JSON Schema - like here https://github.com/asyncapi/asyncapi-react/blob/master/library/src/components/Extensions.tsx#L25

cc @derberg @Min2who @sambhavgupta0705

derberg:
sounds good to me, @sambhavgupta0705 feel free to ask more followup questions

sambhavgupta0705:
@magicmatatjahu is there any previous work or PR from which I can better understand about this

sambhavgupta0705:
just a friendly reminder @magicmatatjahu

sambhavgupta0705:
@magicmatatjahu need more clarification on these two points

[...]

PR #994, opened 2024-04-22
feat: custom extension rendering

ductaily:
**Description**

- Extends the Extension component
- Allows customized rendering of extensions passed as configuration

**Example**

```tsx
import * as React from "react";
import { render } from "react-dom";
import AsyncApiComponent, { ConfigInterface } from "@asyncapi/react-component";
import { ExtensionComponentProps }  from "@asyncapi/react-component/src/components/Extensions"

const schema = `
asyncapi: '2.0.0'
info:
  title: Example
  version: '0.1.0'
channels:
  example-channel:
    subscribe:
      message:
        payload:
          type: object
          x-custom-extension: 'Customized Extension'
          properties:
            exampleField:
              type: string
            exampleNumber:
              type: number
            exampleDate:
              type: string
              format: date-time
`;

function MyCustomExtensionRender(props: ExtensionComponentProps) {
  return (
    <div className="flex py-2">
      <div className="min-w-1/4 mr-2">
        <span className="break-anywhere text-sm italic">
          {props.propertyName}
        </span>
      </div>

      <div className="text-sm prose">
        <a target="_blank" href="#">
          {props.propertyValue}
        </a>
      </div>
    </div>
  )
}

const config: ConfigInterface = {
  schemaID: 'custom-spec',
  show: {
    operations: false,
    errors: false,
  },
  extensions: {
  'x-custom-extension': MyCustomExtensionRender
 }
};

const App = () => <AsyncApiComponent schema={schema} config={config} />;

render(<App />, document.getElementById("root"));
```

Bildschirmfoto 2024-04-22 um 18 44 26

**Related issue(s)**

See also https://github.com/asyncapi/asyncapi-react/issues/819


derberg (commented):
I assume that docs, test and playground showcase is not here yet as you first what to have us look at implementation to give green light for more?

than please at least extend PR with default extensions components that are supported by react component by default? probably best if they are in new dir like `/library/src/components/extensions`. You can take `x-x` and `x-linkedin` as first official extensions under `info` object

magicmatatjahu:
@ductaily Hi! Great work, but you shouldn't override the logic for `Schema` component, but for `Extensions` like in my https://github.com/asyncapi/asyncapi-react/issues/819#issuecomment-1867433195 comment. By this you will be able to add extension not only for `schemas` but also for another sections of spec with possibility to define extensions, like `channels`, `operations` etc.

ductaily:
> @ductaily Hi! Great work, but you shouldn't override the logic for `Schema` component, but for `Extensions` like in my [#819 (comment)](https://github.com/asyncapi/asyncapi-react/issues/819#issuecomment-1867433195) comment. By this you will be able to add extension not only for `schemas` but also for another sections of spec with possibility to define extensions, like `channels`, `operations` etc.

Hi, thank you for reviewing.
We don't understand the difference between our changes and how it would if we made the changes in the Extensions.tsx component.

```tsx
import React, { useContext, useState } from 'react';

import { Schema } from './Schema';

import { SchemaHelpers } from '../helpers';
import { AsyncAPIDocumentInterface } from '@asyncapi/parser';
import { useConfig, useSpec } from '../contexts';
import { CollapseButton } from './CollapseButton';

interface Props {
  name?: string;
  item: any;
}

export interface ExtensionComponentProps<V = any> {
  propertyName: string;
  propertyValue: V;
  document: AsyncAPIDocumentInterface;
}

const SchemaContext = React.createContext({
  reverse: false,
});

export const Extensions: React.FunctionComponent<Props> = ({
  name = 'Extensions',
  item,
}) => {
  const { reverse } = useContext(SchemaContext);
  const [expanded, setExpanded] = useState(false);
  const [deepExpand, setDeepExpand] = useState(false);

  const config = useConfig();
  const document = useSpec();

  const extensions = SchemaHelpers.getCustomExtensions(item);
  if (!extensions || !Object.keys(extensions).length) {
    return null;
  }

  if (!config.extensions || !Object.keys(config.extensions).length) {
    const schema = SchemaHelpers.jsonToSchema(extensions);
    return (
      schema && (
        <div className="mt-2">
          <Schema schemaName={name} schema={schema} onlyTitle={true} />
        </div>
      )
    );
  }

  return (
    <div>
      <div className="flex py-2">
        <div className="min-w-1/4">
          <>
            <CollapseButton
              onClick={() => setExpanded(prev => !prev)}
              expanded={expanded}
            >
              <span className={`break-anywhere text-sm ${name}`}>{name}</span>
            </CollapseButton>
            <button
              type="button"
              onClick={() => setDeepExpand(prev => !prev)}
              className="ml-1 text-sm text-gray-500"
            >
              {deepExpand ? 'Collapse all' : 'Expand all'}
            </button>
          </>
        </div>
      </div>
      <div
        className={`rounded p-4 py-2 border bg-gray-100 ${
          reverse ? 'bg-gray-200' : ''
        } ${expanded ? 'block' : 'hidden'}`}
      >
        {Object.keys(extensions).map(extensionKey => {
          if (config.extensions && config.extensions[extensionKey]) {
            const CustomExtensionComponent = config.extensions[extensionKey];
            return (
              <CustomExtensionComponent
                propertyName={extensionKey}
                propertyValue={extensions[extensionKey]}
                document={document}
              />
            );
          } else {
            const extensionSchema = SchemaHelpers.jsonToSchema(
              extensions[extensionKey],
            );
            return (
              <div className="mt-2">
                <Schema schemaName={extensionKey} schema={extensionSchema} />
              </div>
            );
          }
        })}
      </div>
    </div>
  );
};
```

It would duplicate lots of logic of the Schema component.
Should we have a call?

magicmatatjahu:
@ductaily Hi! From the perspective of code maybe it looks like the code duplication, but extensions are not a "extended" schemas, but separate sections of spec, so even if in code it's better to have new logic inside Schema component, from spec perspective (and user) is better to have separate component.  Also, if you will add logic in Schema component, new logic will be executed for all schemas in spec. Atm we treat extensions as schemas - we don't have logic to render extensions in custom way so it was ok, now it is not. Also, some people uses the Schema component outside the AsyncAPI component in their code, so using `useSpec` and `useConfig` hooks (in my perspective) is not allowed - even if we will use default values for React context.

Next thing, I added (in my comment https://github.com/asyncapi/asyncapi-react/issues/819#issuecomment-1867433195) `parent` prop to the custom extension component not in vain. You can have extension with this same name in two different places, e.g. channel and operation objects and you should be able to write logic base on that info, if you operate on operation extension or channel extension or another one.

Also as I know, you can have a problem with circular references between extension. It's not a problem for extension because if extension allows to have circular refs, then author for custom component should focus on resolving that problem - still if extension won't have a custom component, then Schema component should render it and it should be handled (I see in your new example, and you handled that case). 

A couple of comments for the new code:
- please add that `parent` prop (you can pass `item` prop from ExtensionsProps)
- render all extensions with custom components before all extensions without them, you can even put all rest extensions to single Schema and  render it as:

  ```ts
  const schema = SchemaHelpers.jsonToSchema(restExtensions);
  return (
    schema && (
      <div className="mt-2">
        <Schema schemaName={name} schema={schema} onlyTitle={true} />
      </div>
    )
  );
  ```

  so you will end with less code.
  
If you will have a problems, lets us know!

pavelkornev:
> render all extensions with custom components before all extensions without them, you can even put all rest extensions to single Schema and render it as:

I would oppose to this requirement. It's better to sort extensions by name (alphabetically) and render in that order. In our use case (SAP) we have some fields with shared prefixes. Some of those fields will have a custom component and some won't. It would mean they will be rendered apart from each other.

**Example:**

- x-sap-odm-entity-name (for this field we plan to have a custom component with a link to elsewhere from the field value)
- x-sap-odm-oid (no need for a custom component)

**Expected behaviour:** both fields are rendered close to each other in UI.

magicmatatjahu:
@pavelkornev HI! Sorry for delay!

> I would oppose to this requirement. It's better to sort extensions by name (alphabetically) and render in that order. In our use case (SAP) we have some fields with shared prefixes. Some of those fields will have a custom component and some won't. It would mean they will be rendered apart from each other.

Ok, I understand :)

library/src/components/Extensions.tsx
```diff
@@ -1,34 +1,112 @@
...
+export interface ExtensionComponentProps<V = any> {
+  propertyName: string;
+  propertyValue: V;
+  document: AsyncAPIDocumentInterface;
+  parent: BaseModel;
+}
+
+const SchemaContext = React.createContext({
+  reverse: false,
+});
```
magicmatatjahu, line 28: You don't use that context in code (only read the context value), so it can be removed.

```suggestion
```

library/src/components/Extensions.tsx
```diff
@@ -1,34 +1,112 @@
...
+
 export const Extensions: React.FunctionComponent<Props> = ({
   name = 'Extensions',
   item,
 }) => {
+  const { reverse } = useContext(SchemaContext);
```
magicmatatjahu, line 34: See my previous comment about context.

```suggestion
```

library/src/components/Extensions.tsx
```diff
@@ -1,34 +1,112 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
 
-import React from 'react';
+import React, { useContext, useState } from 'react';
```
magicmatatjahu, line 4: See my previous comment about context.

```suggestion
import React, { useState } from 'react';
```

magicmatatjahu (changes requested):
Some things what I noticed. I have to also run that code locally and check some edge cases. Please update branch with latest changes. I hope it won't be a problem if I will have to change slighty the code in branch? But overall, great job! :)

ductaily:
> I assume that docs, test and playground showcase is not here yet as you first what to have us look at implementation to give green light for more?
> 
> than please at least extend PR with default extensions components that are supported by react component by default? probably best if they are in new dir like `/library/src/components/extensions`. You can take `x-x` and `x-linkedin` as first official extensions under `info` object

@derberg I will add the docs, test and playground showcase.

Regarding your second comment just to clarify.
If I understood you correctly, you want to add in `library/src/config/default.ts` the following:

```ts
 ...
  extensions: {
    'x-x': DefaultExtensionComponent,
    'x-linkedin': DefaultExtensionComponent,
  },
```

Having something like `library/src/components/supportedExtensionComponents/DefaultExtensionComponent.tsx` with the content:

```tsx
export default function DefaultExtensionComponent(
  props: ExtensionComponentProps,
) {
  return (
    <div className="mt-2">
      <div className="flex py-2">
        <div className="min-w-1/4 mr-2">
          <span className="break-anywhere text-sm italic">
            {props.propertyName}
          </span>
        </div>
        <div>
          <div className="text-sm prose">{props.propertyValue}</div>
        </div>
      </div>
    </div>
  );
}
```

Is that correct?

derberg:
@ductaily not default component, but specific component for each extension, just like you will have probably some specific component for `x-sap-odm-oid`

so in case of `x-x` there is a component that renders X logo that is a clickable link pointing to twitter. Same with LinkedIn.

so they are default in a way they are included in the library, but still custom, if you know what I mean. 

great is that they will be a nice, production showcase how to add extensions

pavelkornev:
@derberg you mean to keep all extension components within this repo? We have use cases where we would need to set internal links, therefore we cannot commit such components to this public repo. I think both use cases can co-exist.

derberg:
@ductaily what node version are you using?

@kennethaasan can you have a look?

derberg:
I did
```
npm run clean
npm install
npm run start
```

on node 20 on master and playground renders well

when I switch to your branch, installation fails witj
```
npm ERR! code 1
npm ERR! path /Users/wookiee/sources/asyncapi-react/library
npm ERR! command failed
npm ERR! command sh -c npm run build:dev
npm ERR! > @asyncapi/react-component@2.0.0 build:dev
npm ERR! > npm run build:esm && npm run build:types && npm run build:styles
npm ERR! 
npm ERR! 
npm ERR! > @asyncapi/react-component@2.0.0 build:esm
npm ERR! > tsc -p tsconfig.esm.json
npm ERR! 
npm ERR! src/components/Extensions.tsx(76,11): error TS2304: Cannot find name 'reverse'.
npm ERR! npm ERR! Lifecycle script `build:esm` failed with error: 
npm ERR! npm ERR! Error: command failed 
npm ERR! npm ERR!   in workspace: @asyncapi/react-component@2.0.0 
npm ERR! npm ERR!   at location: /Users/wookiee/sources/asyncapi-react/library 
npm ERR! npm ERR! Lifecycle script `build:dev` failed with error: 
npm ERR! npm ERR! Error: command failed 
npm ERR! npm ERR!   in workspace: @asyncapi/react-component@2.0.0 
npm ERR! npm ERR!   at location: /Users/wookiee/sources/asyncapi-react/library

npm ERR! A complete log of this run can be found in: /Users/wookiee/.npm/_logs/2024-05-13T10_18_29_742Z-debug-0.log
```

ductaily:
Hi @derberg , 

Thank you for checking.
I updated the `Extensions.tsx`. 
You should be able to install it now.

kennethaasan:
Still need me to check? I've been offline on holiday so have not seen this until now

ductaily:
> Still need me to check? I've been offline on holiday so have not seen this until now

> @ductaily after your recent update, installation works, and playground renders as expected - not sure what should I do to replicate your issue

Hi @derberg, @kennethaasan,
sorry maybe my explanation was lacking.

Could you please check the following:
1. do local changes in one of the `/library/**/*.tsx`
2.  run start
Do you see those local changes rendered? The development environment does not recognize my changes.

kennethaasan:
You're right, it stopped working. I'm pretty sure I checked this when we migrated from lerna to npm workspaces, but I guess some old linking was still in place on my machine. It looks like we need to change the `playground/package.json` file to be `"@asyncapi/react-component": "file:../library"` for npm to link it correctly on install. But I can't do that because there's a release workflow which changes this version. @derberg, perhaps we can get rid of the `bump-react-comp.sh` scripts and instead let npm resolve the version. Please see https://www.linkedin.com/pulse/things-i-wish-had-known-when-started-javascript-monorepo-gorej/. This should simplify things quite a bit in the repo.

derberg:
@kennethaasan you're completely right. Tbh I though that for playground we did it already as anyway playground is not released to npm. Btw, @char0n that wrote the article that you linked, is maintainer of asyncapi spec and also uses this react component in swagger editor 😄 small world 😄 

@ductaily please do the following:
- remove this step https://github.com/asyncapi/asyncapi-react/blob/master/.github/workflows/release-wc-and-playground.yml#L79-L81
- remove `npm run install:reactcomp:playground` from https://github.com/asyncapi/asyncapi-react/blob/master/package.json#L41
- remove this script https://github.com/asyncapi/asyncapi-react/blob/master/package.json#L42
- remove https://github.com/asyncapi/asyncapi-react/blob/master/playground/package.json#L8
- remove that file https://github.com/asyncapi/asyncapi-react/blob/master/playground/bump-react-comp.sh
- and make a change in package.json, about file reference that @kennethaasan described in his comment

kennethaasan:
haha, very small world. And thanks for fixing this issue!

ductaily:
@derberg @kennethaasan Thanks a lot for your help :)


1. I added an example for the `x-x` extension to render a clickable logo. @derberg is this how you imagined it? (See `library/src/config/default.ts` and `library/src/components/supportedExtensions/XExtension.tsx`)
2. I also added documentation to `docs/configuration/config-modification.md`. Is there any other place where I should add it?
3. I am unsure how the changes for the Playground showcases should look like. Could you please guide me here?

derberg:
sorry for being late, we had some issues. Me and few other maintainers lost funding https://www.asyncapi.com/blog/2024-june-summary#postman-partnership-changes. Please join AsyncAPI Slack where you can ping me when you see an unusual delay on review

> I am unsure how the changes for the Playground showcases should look like. Could you please guide me here

just add example `x-x: AsyncAPISpec` to https://github.com/asyncapi/asyncapi-react/blob/master/playground/specs/streetlights.ts so when playground is loaded, the example is showcased there and visible

also, once you do it, please add a screen shot to show where it can be found. I have to admit I quickly added it and could not locate where it is rendered

ductaily:
> sorry for being late, we had some issues. Me and few other maintainers lost funding https://www.asyncapi.com/blog/2024-june-summary#postman-partnership-changes. Please join AsyncAPI Slack where you can ping me when you see an unusual delay on review
> 
> > I am unsure how the changes for the Playground showcases should look like. Could you please guide me here
> 
> just add example `x-x: AsyncAPISpec` to https://github.com/asyncapi/asyncapi-react/blob/master/playground/specs/streetlights.ts so when playground is loaded, the example is showcased there and visible
> 
> also, once you do it, please add a screen shot to show where it can be found. I have to admit I quickly added it and could not locate where it is rendered

Hey @derberg, thank you for the guidance :)
I added an example:

image

derberg (commented):
the only problem is that `x-x` extension should be used only in `info` object level, not channel

https://github.com/asyncapi/extensions-catalog/blob/master/extensions/x.md

@magicmatatjahu can you also have a look please

pavelkornev:
> the only problem is that `x-x` extension should be used only in `info` object level, not channel
> 
> https://github.com/asyncapi/extensions-catalog/blob/master/extensions/x.md
> 
> @magicmatatjahu can you also have a look please

Custom extensions from Info block are not rendered by UI at all. We can replace the example with some other custom field. Which field can you recommend to use instead as an example?

library/src/components/Extensions.tsx
```diff
@@ -1,34 +1,107 @@
...
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export interface ExtensionComponentProps<V = any> {
+  propertyName: string;
+  propertyValue: V;
+  document: AsyncAPIDocumentInterface;
+  parent: BaseModel;
+}
+
```
AceTheCreator, line 26: You should probably move this to the types file.

ductaily: I moved it into the types file

library/src/components/Extensions.tsx
```diff
@@ -1,34 +1,107 @@
...
 export const Extensions: React.FunctionComponent<Props> = ({
   name = 'Extensions',
   item,
 }) => {
+  const [expanded, setExpanded] = useState(false);
+  const [deepExpand, setDeepExpand] = useState(false);
```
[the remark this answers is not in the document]
ductaily: I agree with you. I removed `deepExpand`.

library/src/components/supportedExtensions/XExtension.tsx
```diff
@@ -0,0 +1,36 @@
...
+        viewBox="0 0 1200 1227"
+        fill="none"
+        xmlns="http://www.w3.org/2000/svg"
+      >
+        <path
+          d="M714.163 519.284L1160.89 0H1055.03L667.137 450.887L357.328 0H0L468.492 681.821L0 1226.37H105.866L515.491 750.218L842.672 1226.37H1200L714.137 519.284H714.163ZM569.165 687.828L521.697 619.934L144.011 79.6944H306.615L611.412 515.685L658.88 583.579L1055.08 1150.3H892.476L569.165 687.854V687.828Z"
+          fill="black"
+        />
+      </svg>
+    </div>
```
AceTheCreator, line 34: Out of curiosity why aren't we wrapping it an anchor tag?

library/src/components/supportedExtensions/XExtension.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { ExtensionComponentProps } from '../../types';
+
+/**
+ * See <https://github.com/asyncapi/extensions-catalog/blob/master/extensions/x.md>.
+ */
+export default function XExtension({
+  propertyValue,
+}: ExtensionComponentProps<string>) {
+  const onClickHandler = () => {
+    window.open(`https://x.com/${propertyValue}`, '_blank');
```
AceTheCreator, line 12: I think for best practices we should include the `noopener,noreferrer` attribute

ductaily: I added the attribute in the anchor tag.

library/src/components/supportedExtensions/XExtension.tsx
ductaily: Thank you for pointing it out. I wasn't paying attention here :)

AceTheCreator (approved):
LGTM! 👌🏽

ductaily:
@AceTheCreator @magicmatatjahu Is there anything I should do for the PR to be merged?
The auto merge seems to be blocked because of the change requested but this should already be addressed.

pavelkornev:
@magicmatatjahu could you please approve this PR as a code owner?

magicmatatjahu (approved):
LGTM

Issue #891, opened 2024-01-24
update project dependencies


vishvamsinh28:
This issue aims to update the project dependencies, encompassing the latest Node.js version, npm or Yarn updates, and the upgrading of project dependencies to their latest compatible versions.

kennethaasan:
Just realized that this project needs Node v16 which is now dead. It fails when you try to start it with Node v18. We need this project to support >= v18

kennethaasan:
- [ ] migrate from react-scripts to next.js: #929 
- [ ] migrate from tslint to eslint
- [ ] upgrade react.js to v18 in the library
- [ ] upgrade dependencies

derberg:
reopenning
looks like this issue was marked as being resolved with playground PR - but that is not the case yet

derberg:
@kennethaasan will you proceed with other tasks?

kennethaasan:
Yes, that's the plan. I'll get to it in April/May.

PR #996, opened 2024-04-23
refactor: migrate from tslint to eslint

kennethaasan:
**Description**

- migrate from tslint to eslint

**Related issue(s)**

See also #891


kennethaasan:
@derberg, please review. I know it's a lot of changes, but it's kind of necessary in this PR. I added some strict TS rules because this repo has way too much use of any. That should be the exception and not the norm.

derberg:
lgtm, just thinking if we really need a release, `refactor: ` should be fine imho

kennethaasan:
@derberg, I was unsure myself. Let's do refactor

derberg (approved):

PR #997, opened 2024-04-24
refactor: upgrade prettier and format code accordingly

kennethaasan:
**Description**

- upgrade prettier and format code accordingly

**Related issue(s)**

See also #891


derberg (approved):
@kennethaasan yeah, it is the same with eslint. There are ideas, just noone here that can drive cross project initiatives. Try to reach `wg-maintainers-growth` in slack, maybe this is a topic for them - maybe unification of these lowers maintenance complexity

PR #1019, opened 2024-06-24
fix: remove code needed prior react 18 to fix release of webcomponent

derberg:
webcomponent and playground releases were failing

https://github.com/asyncapi/asyncapi-react/actions/runs/9496432356/job/26170818500

I also added now slack notifications for failing releases as we had no idea and nobody explicitly wrote to us about it

I learned from https://github.com/asyncapi/asyncapi-react/pull/1016#issuecomment-2186453509


magicmatatjahu (approved):
LGTM!

PR #1021, opened 2024-06-25
fix: update web-component bumping script for react

derberg:
Looks like depending on npm versions there are different behaviours of installation

even though existing script has explicit `--save`, if some other script in the same run updated dependency to lates, and it is already in node_modules, installation of the same dependency with `--save` in another project in monorepo, will not result in updates in package.json

This this PR introduce initial remove of dependency and then fresh installation


derberg:
@magicmatatjahu it fails because of package-lock caused by the issue that this PR solves. Once you approve I will merge as admin

magicmatatjahu (approved):

PR #1023, opened 2024-06-26
fix: render specs that contain oauth2 password flow only

### LINKED ISSUE #1007: [BUG] Fails to render the specification if it contains oauth2 security scheme with password flow
**Description**

When the specification contains securityScheme of type oauth2 that contains password flow and that scheme is referred in the document the spec fails to render. It happens in all the major browsers.

**Expected result**

The specification must be rendered successfully.

**Actual result**

The specification is not rendered, we get blank screen instead.

**Steps to reproduce**

The fastest way to reproduce is in async api studio which uses this component under the hood.
Add the following under components/securitySchemes:

```
    oauth2Password:
      type: oauth2
      description: OAuth 2.0 example
      flows:
        password:
          tokenUrl: https://auth.example.com/oauth/token
          refreshUrl: https://auth.example.com
          availableScopes:
            read:user: Read user information
```

Just adding the scheme is ok but if we further refer the scheme it will result in browser rendering a blank page. The error that can be seen in the console is:

Screenshot 2024-05-29 at 10 54 39

**Troubleshooting**

[As described in the spec](https://www.asyncapi.com/docs/reference/specification/v3.0.0#oauthFlowObject) the authorizationUrl does not apply to password flow.

However [in the code it is handled](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/containers/Servers/Security.tsx#L171) as always present, which causes "undefined is not an object (evaluating 'i.authorizationUrl')" error.

pltod2:
**Description**

Fix for rendering specifications that contain oauth2 password flow only.

**Related issue(s)**

Resolves #1007


pltod2:
@derberg  Do you have any idea when this PR can be looked at and merged? Thanks for your response!

derberg (commented):
can you extend https://github.com/asyncapi/asyncapi-react/blob/master/library/src/__tests__/docs/v3/streetlights-mqtt.json instead of creating a new test document?

pltod2:
@derberg done

derberg (commented):
sorry, I meant like edit the existing YAML file and add additional configuration that you can use in test.

we always try to make sure that whatever new case we support and test, is the easily "quick to see" in https://asyncapi.github.io/asyncapi-react/

derberg (approved):

derberg:
Thanks for the PR!

/rtm

PR #1034, opened 2024-08-06
fix: update isomorphic-dompurify and remove @types/highlight.js

batyshkaLenin:

- Remove @types/highlight.js, because highlight.js have own type definitions
- Update isomorphic-dompurify and @types/dompurify

### DIFF
--- a/library/package.json
+++ b/library/package.json
@@ -73,7 +73,7 @@
     "@asyncapi/parser": "^3.1.0",
     "@asyncapi/protobuf-schema-parser": "^3.2.14",
     "highlight.js": "^10.7.2",
-    "isomorphic-dompurify": "^2.12.0",
+    "isomorphic-dompurify": "^2.14.0",
     "marked": "^4.0.14",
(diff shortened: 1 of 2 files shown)


derberg:
@batyshkaLenin thanks, can you share why these updates : `Update isomorphic-dompurify and @types/dompurify` are needed?

batyshkaLenin:
@derberg @types/highlight.js I removed because this library has its own type declarations. isomorphic-dompurify I updated because, every time I installed dependencies in my project, there was a warning that one of its dependencies had a CVE (I don't remember which one, I made this PR over two weeks ago). @types/dompurify I updated to not have conflicting type versions in package-json, because isomorphic-dompurify has a different type version. In a broader view, there was no big need to upgrade, I wanted to get rid of the warning when installing dependencies in my project.

derberg (approved):
Thanks, since we update dev dependency, we should release patch, this is why I changed the prefix to `fix:`

Issue #962, opened 2024-03-26
508 Accessibility: Opening content in a new window


ghost:
We use a tool called Sort Site to evaluate our site for Section 508 Accessibility.  It is flagging that links to content either should not open in a new window or else should inform users that the content will be opened in a new window.  The error information is:  

![image](https://github.com/asyncapi/asyncapi-react/assets/86489380/7e92efc9-08ba-4179-bb20-8c6ec2437177)

The source of these errors on our page is the APPLICATION/XML and EXTERNAL DOCS links:
![image](https://github.com/asyncapi/asyncapi-react/assets/86489380/a1c3fde1-8f28-4623-8c4d-4f2c83dcc579)

Either these links should display the content in the same window, or there should be text like "(opens in a new window)" with the links.

aakankshabhende:
@ghost I would like to work on this issue. Could you please assign me this?

AceTheCreator:
@aakankshabhende, we don't assign issues, feel free to pick it up :)

PR #1060, opened 2024-10-02
fix: external link will open in the same window

ItshMoh:
Now the external link will open in the same window.

**Description**

**Related issue(s)**
#962 
[video](https://www.loom.com/share/a4d6280fb6564479b6204c12c0fcf2ea?sid=5f0cc668-4335-431d-966e-4d6b7a2ccee2)

### DIFF
--- a/library/src/components/Href.tsx
+++ b/library/src/components/Href.tsx
@@ -17,7 +17,6 @@ export const Href: React.FunctionComponent<Props> = ({
     href={href}
     title={title}
     className={className}
-    target="_blank"
     rel="nofollow noopener noreferrer"
   >
     {children}
--- a/library/src/components/Href.tsx
+++ b/library/src/components/Href.tsx
@@ -15,8 +15,9 @@ export const Href: React.FunctionComponent<Props> = ({
 }) => (
   <a
     href={href}
-    title={title}
+    title={title ? `${title} (Opens in new window)` : 'Opens in new window'}
     className={className}
+    target="_blank"
     rel="nofollow noopener noreferrer"
   >
     {children}


DhairyaMajmudar (approved):
Looks Good 🚀

AceTheCreator:
@ItshMoh, i don't think external link should open in the same window, instead we should let users know it opens a new window as indicated in the issue :)

AceTheCreator (changes requested):
@ItshMoh, i don't think external link should open in the same window, instead we should let users know it opens a new window as indicated in the issue :)

ItshMoh:
> @ItshMoh, i don't think external link should open in the same window, instead we should let users know it opens a new window as indicated in the issue :)

Hey @AceTheCreator Do you want the message for opening to the new window like this. 
alert_new_window_react
Here when user clicks the link it will get alert notifcations.

AceTheCreator:
> > @ItshMoh, i don't think external link should open in the same window, instead we should let users know it opens a new window as indicated in the issue :)
> 
> Hey @AceTheCreator Do you want the message for opening to the new window like this. alert_new_window_react Here when user clicks the link it will get alert notifcations.

@ItshMoh, it doesn't have to be an alert, you can just add the warning message inside the title attribute like this:

<a href="https://example.com" title="Warning">Visit Example Site</a>

The warning will appear as tooltip when link is hovered over.

ItshMoh:
> > > @ItshMoh, i don't think external link should open in the same window, instead we should let users know it opens a new window as indicated in the issue :)
> > 
> > 
> > Hey @AceTheCreator Do you want the message for opening to the new window like this. alert_new_window_react Here when user clicks the link it will get alert notifcations.
> 
> @ItshMoh, it doesn't have to be an alert, you can just add the warning message inside the title attribute like this:
> 
> [Visit Example Site](https://example.com)
> 
> The warning will appear as tooltip when link is hovered over.

Hey @AceTheCreator  i have added the tooltip on hovering over the button. 
react_open_new_window

AceTheCreator (approved):
LGTM!

PR #1063, opened 2024-10-02
refactor: types refactoring

DhairyaMajmudar:
**Description**

-> Refactoring files by removing any keywords and typescript warnings disabling comments 

cc: @AceTheCreator 

**Related issue(s)**
N/A

### DIFF
--- a/library/src/components/Extensions.tsx
+++ b/library/src/components/Extensions.tsx
@@ -1,16 +1,13 @@
-/* eslint-disable @typescript-eslint/no-unsafe-return */
-/* eslint-disable @typescript-eslint/no-unsafe-assignment */
-
 import React from 'react';
 
 import { Schema } from './Schema';
+import type { SchemaInterface } from '@asyncapi/parser';
 
 import { SchemaHelpers } from '../helpers';
 
 interface Props {
   name?: string;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  item: any;
+  item: SchemaInterface;
 }
 
 export const Extensions: React.FunctionComponent<Props> = ({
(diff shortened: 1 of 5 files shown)


AceTheCreator (changes requested):
Kindly look into why your build is failing :/

DhairyaMajmudar:
> Kindly look into why your build is failing :/

I've fixed them 👍🏻

web-component/src/AsyncApiWebComponent.tsx
```diff
@@ -50,7 +48,7 @@ function retrieveSchemaProp(
 export interface AsyncApiWebComponentProps extends AsyncApiProps {
   cssImportPath?: string;
   schemaFetchOptions?: RequestInit;
-  schemaUrl: string;
+  schemaUrl?: string;
```
AceTheCreator, line 51: @DhairyaMajmudar the `schemaUrl` is not an optional value

library/src/contexts/useSpec.ts
```diff
@@ -2,8 +2,7 @@ import React, { useContext } from 'react';
 import { AsyncAPIDocumentInterface } from '@asyncapi/parser';
 
 export const SpecificationContext =
-  // eslint-disable-next-line @typescript-eslint/no-unsafe-argument, @typescript-eslint/no-explicit-any
-  React.createContext<AsyncAPIDocumentInterface>(null as any);
+  React.createContext<AsyncAPIDocumentInterface>(null as never);
```
AceTheCreator, line 5: I think your changes here is unnecessarily strict and semantically incorrect in this scenario. I'd suggest we use the existing type :)

library/src/components/Extensions.tsx
```diff
@@ -1,6 +1,5 @@
 /* eslint-disable @typescript-eslint/no-unsafe-return */
 /* eslint-disable @typescript-eslint/no-unsafe-assignment */
-
```
AceTheCreator, line 3: I think this line should be removed

AceTheCreator (changes requested):
Left some reviews for you

Issue #958, opened 2024-03-25
508 Accessibility: Button must have a focus outline


ghost:
We are using Sort Stie to evaluate our site for Section 508 Accessibility compliance.  It is flagging a button does not have a focus outline.

![image](https://github.com/asyncapi/asyncapi-react/assets/86489380/424b7dcb-e256-4515-8dd1-968fc7b08816)

This "Server specific information" is the violating component; I didn't even realize this was a button if it weren't for the Sort Site error!  Seems there is no need to have this be a button since there is the Expand All component?
![image](https://github.com/asyncapi/asyncapi-react/assets/86489380/c3ba54be-f209-48ef-ba1c-811533e21eea)

Aditya90456:
i can fix this

Orthodox-64:
can u assign this issue to me @AceTheCreator

blazethunderstorm:
@AceTheCreator is this issue open to work ?

AadarshM07:
@AceTheCreator , I would like to work on this issue. Can you assign this to me? Thank you!

PR #1065, opened 2024-10-03
fix: removed the focus: outline-none and added a visible focus.

ItshMoh:
Removed the focus: outline-none and added a visible focus to the buttons.

**Description**

**Related issue(s)**
#958
button_focus

### DIFF
--- a/library/src/components/CollapseButton.tsx
+++ b/library/src/components/CollapseButton.tsx
@@ -33,7 +33,7 @@ export const CollapseButton: React.FunctionComponent<Props> = ({
 }) => (
   <button
     {...rest}
-    className={`focus:ring-2 focus:ring-blue-500 focus:ring-offset-2 ${rest.className}`}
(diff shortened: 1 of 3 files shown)


library/src/components/CollapseButton.tsx
```diff
@@ -33,7 +33,7 @@ export const CollapseButton: React.FunctionComponent<Props> = ({
 }) => (
   <button
     {...rest}
-    className={`focus:outline-none ${rest.className}`}
+    className={`focus:ring-2 focus:ring-blue-500 focus:ring-offset-2 ${rest.className}`}
```
DhairyaMajmudar, line 36: Here instead of using `blue-500` color variant how's about using something which matches with the existing theme ?

AceTheCreator (changes requested):
Please use a color that matches with the existing theme

Issue #1069, opened 2024-10-06
Cant switch tabs multiple time. It gets stuck on "CONFIGURATION"


catosaurusrex2003:
**Description**
Current Behaviour
If you go from `SCHEMA` tab to the `CONFIGURATION`, then you cant go back.

**Actual result**

https://github.com/user-attachments/assets/30bba160-742a-458a-bd4f-7d6896c55c8f

**Expected result**

https://github.com/user-attachments/assets/1e8d9ec3-e37a-43f4-ae52-a2abb2926b2b

**Steps to reproduce**
[https://asyncapi.github.io/asyncapi-react/](https://asyncapi.github.io/asyncapi-react/) go here and click on `CONFIGURATION` then you cant go back to `SCHEMA`

ItshMoh:
Hey @catosaurusrex2003 this issue is already been reported. #1008 .

catosaurusrex2003:
My bad. I couldn't find it when i searched for similar issues.

PR #1070, opened 2024-10-06
fix: ensure tabIndex is a number before proceeding in condition

catosaurusrex2003:
**Description**
Checking tabIndex is a number instead of checking if tabIndex is falsy or truthy  

**Related issue(s)**
#1069


DhairyaMajmudar (approved):
Working Good 👍🏻

PR #1090, opened 2024-10-15
feat : Add description tooltips to tags pills

Manancode:
I've created a reusable TooltipWrapper component in library/src/components/TooltipWrapper.tsx
However, I've been unable to locate the exact component or file where the External Docs information is currently rendered. As a result, I can't yet implement the full solution. It would be helpful if someone helps me with the location of external docs information . Till then I am submitting this draft PR which includes a TooltipWrapper component.

### DIFF
--- a/library/src/components/ToolTipWrapper.tsx
+++ b/library/src/components/ToolTipWrapper.tsx
@@ -1,0 +1,32 @@
+import React, { ReactNode } from 'react';
+import * as RadixTooltip from '@radix-ui/react-tooltip';
+
+interface TooltipWrapperProps {
+  children: ReactNode;
+  description?: string;
+}
+
+export const TooltipWrapper: React.FC<TooltipWrapperProps> = ({ children, description }) => {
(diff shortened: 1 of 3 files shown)


catosaurusrex2003:
Hey @Manancode 

i would suggest 

instead of adding any new dependencies for tootips

we can use the native tool-tips `title` of html
just like how it has been done in [tags](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/components/Tag.tsx#L26)
here in [https://github.com/asyncapi/asyncapi-react/blob/master/library/src/components/Href.tsx#L18](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/components/Href.tsx#L18)

also
rename your pr from `feat : {your text}` to `feat: {your text}` . we follow [https://www.conventionalcommits.org/en/v1.0.0/](https://www.conventionalcommits.org/en/v1.0.0/) and hence your Pr linting is failing.

PR #1093, opened 2024-10-16
feat: add TooltipWrapper component for enhanced UI tooltips.

Manancode:
This Pull Request introduces the TooltipWrapper component, which provides a reusable solution for displaying tooltips in the UI. The component accepts three props:

> children: The elements to be wrapped and displayed.
> description (optional): A string that serves as the tooltip text.
> className (optional): A string for custom styling of the tooltip wrapper

If the description prop is not provided, the component renders the children without any additional markup.

To use this wrapper :
import { TooltipWrapper } from "./components/ToolTipWrapper";
      <TooltipWrapper description="Another tooltip" className="custom-class">
        <span>This also has a tooltip</span>
      </TooltipWrapper>


Manancode:
I've updated the PR based on feedback. I've removed the TooltipWrapper component and instead added a title attribute directly to the externalDocs link in Info.tsx. This will show the externalDocs description as a tooltip when hovering over the link. Please review and let me know if any further changes are needed. @catosaurusrex2003

catosaurusrex2003:
Great job @Manancode :rocket:

Everything is perfect. I just think that displaying EXTERAL_DOCUMENTATION_TEXT when there is no description will be repetitive.

After that is fixed, it is rtm @AceTheCreator

Manancode:
Done sir @catosaurusrex2003 . Added undefined so the tooltip will only appear when there's a description available. If there's no description, no tooltip will be shown, preventing the repetition of the link text. 
Without undefined, an empty tooltip (usually just a small, blank box) would appear when hovering over the link if there's no description.

catosaurusrex2003:
perfectly done 👌

LGTM :100:

AceTheCreator (changes requested):
@Manancode kindly look into why your lint check is failing

Manancode:
Done sir @AceTheCreator

AceTheCreator (approved):
LGTM!

Issue #1100, opened 2024-10-22
unrecoverable error in case of recursion in asyncapi document


catosaurusrex2003:
**Description**
As described in this issue https://github.com/asyncapi/studio/issues/771

when a document has got recursion in it

the rendering crashes due to too many recursion.

This causes the app to go into unrecoverable error.

**Steps to reproduce:**
1. go to [https://studio.asyncapi.com](https://studio.asyncapi.com/)
2.  paste
```
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  user/signedup:
    address: user/signedup
    messages:
      subscribe.message:
        $ref: '#/components/messages/TradeMessageServer'

components:
  messages:
    TradeMessageServer:
      name: Trade Message (server)
      title: Trade Message (server)
      description: A message containing trade data
      x-something: 
        $ref: '#/components/messages/TradeMessageServer' #RECURSION IS HAPPENING HERE !!! 
```
3. 
![image](https://github.com/user-attachments/assets/6b6eed06-72b0-439e-99ce-db547370d11d)
the error in console
![image](https://github.com/user-attachments/assets/010a267b-fcc4-43c5-896f-6e1d924c3e72)

Now Since studio fetches the document from localstorage before loading. Even if you reload the site it keeps on crashing again and again. Only way to recover is to clear the localstorage and start again with all the progress lost. 

**Expected result**
Application should not crash into an unrecoverable state.

The problem is in the recursion found in this [function](https://github.com/asyncapi/asyncapi-react/blob/4253f295a2a4bb753f616d2d54ccfac41048ab5d/library/src/helpers/schema.ts#L521-L564) which doesnt have a recursion limit

**Related Issues:**
https://github.com/asyncapi/studio/issues/771

reachaadrika:
@AceTheCreator can I pick this for Level 1 Review ?

catosaurusrex2003:
the Prs #1108 and #1101 are very related

#1108 is responsible for actually rendering the error message. But for the internal errors the message might differ from browser to browser

#1101 is responsible to detect recursion using weakset and throw a more helpful error message.

PR #1101, opened 2024-10-22
fix: application crash by clipping the recursion using WeakSet

catosaurusrex2003:
**Description:**
This fixes the issue of website crashing due to too many recursions.

BUT

Even if we put a recursion limit of 10 to the function like this -> [https://github.com/catosaurusrex2003/asyncapi-react/blob/9b73e41b7a644992851068d6cf327a5775eb7884/library/src/helpers/schema.ts#L521-L569](https://github.com/catosaurusrex2003/asyncapi-react/blob/9b73e41b7a644992851068d6cf327a5775eb7884/library/src/helpers/schema.ts#L521-L569)

We will still have the recursive part rendering 10 times like this
image

The best approach would be to create an `ErrorBoundary` ( [using this](https://github.com/bvaughn/react-error-boundary) ) on the whole application in `library/src/containers/AsyncApi/Layout.tsx`.

So we can throw error indiscriminately instead of a empty object in `jsonFieldToSchema` function. 
We can render the error and make the app recoverable.

Let me know if my current approach is fine or should i modify my PR with the approach with  an `ErrorBoundary`.

**Related issue(s)**
#1100

### DIFF
--- a/.github/workflows/update-maintainers-trigger.yaml
+++ b/.github/workflows/update-maintainers-trigger.yaml
@@ -1,0 +1,28 @@
+# This action is centrally managed in https://github.com/asyncapi/.github/
+# Don't make changes to this file in this repo as they will be overwritten with changes made to the same file in above mentioned repo
+
+name: Trigger MAINTAINERS.yaml file update
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      # Check all valid CODEOWNERS locations: 
+      # https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners#codeowners-file-location
+      - 'CODEOWNERS'
+      - '.github/CODEOWNERS'
+      - '.docs/CODEOWNERS'
+
+jobs:
+  trigger-maintainers-update:
+    name: Trigger updating MAINTAINERS.yaml because of CODEOWNERS change
+    runs-on: ubuntu-latest
+    
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@ff45666b9427631e3450c54a1bcbee4d9ff4d7c0 # https://github.com/peter-evans/repository-dispatch/releases/tag/v3.0.0
+        with:
+          # The PAT with the 'public_repo' scope is required
+          token: ${{ secrets.GH_TOKEN }}
+          repository: ${{ github.repository_owner }}/community
+          event-type: trigger-maintainers-update
--- a/library/package.json
+++ b/library/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/react-component",
-  "version": "2.4.3",
+  "version": "2.5.0",
   "private": false,
   "description": "A React component for AsyncAPI specification.",
   "repository": {
@@ -71,11 +71,12 @@
     "@asyncapi/avro-schema-parser": "^3.0.24",
     "@asyncapi/openapi-schema-parser": "^3.0.24",
     "@asyncapi/parser": "^3.3.0",
-    "@asyncapi/protobuf-schema-parser": "^3.2.14",
+    "@asyncapi/protobuf-schema-parser": "^3.3.0",
     "highlight.js": "^10.7.2",
     "isomorphic-dompurify": "^2.14.0",
     "marked": "^4.0.14",
     "openapi-sampler": "^1.2.1",
+    "react-error-boundary": "^4.1.2",
     "use-resize-observer": "^9.1.0"
   },
   "peerDependencies": {
--- a/library/package.json
+++ b/library/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@asyncapi/react-component",
-  "version": "2.4.2",
+  "version": "2.4.3",
   "private": false,
   "description": "A React component for AsyncAPI specification.",
   "repository": {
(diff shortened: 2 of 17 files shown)


reachaadrika:
@AceTheCreator  can I pick this for Level 1 Review ?

AceTheCreator:
> @AceTheCreator can I pick this for Level 1 Review ?

Please go ahead @reachaadrika

reachaadrika:
@catosaurusrex2003  Thanks for picking this issue .  

Also here I would suggest to go ahead with` ErrorBoundary` , since implement error boundary will be an asset for entire application and won't be limited to this issue itself . 

Also to fix the multiple recursions , rather than specifying the MAX DEPTH to 10 , where we might encounter issue you mentioned above , we can maybe explore `WeakSet ` , 
`A WeakSet can help keep track of objects that have already been processed in the current path of recursion. This avoids re-processing objects and naturally handles circular references without limiting the depth. This approach works well for complex, deeply nested structures, as it relies on detecting cycles rather than counting recursion levels.` 

FYR : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakSet 

PS : Open to discussion / suggestions on the approach , from both you and @AceTheCreator  . 

cc : @AceTheCreator

reachaadrika:
@catosaurusrex2003  Will review this change , once . 
Also please add a screen recording . 

Add yes adding Error Boundary , Would be beneficial in the long run . Let's keep this isolated you can create a new issue for the same as Improvement .

reachaadrika (commented):
Let's test with 1-2 samples , Rest LGTM ✌🏻
@AceTheCreator  Level -1 Review and Testing is done , Please review once too . 
Thanks

catosaurusrex2003:
> Let's test with 1-2 samples

while testing with the following yaml

```yml
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
  description: This service is in charge of processing user signups
channels:
  userSignedup:
    address: user/signedup
    messages:
      UserSignedUp:
        $ref: '#/components/messages/UserSignedUp'
    x-recursion:  #this RECURSION clips correctly
      $ref: '#'
operations:
  sendUserSignedup:
    action: send
    channel:
      $ref: '#/channels/userSignedup'
    messages:
      - $ref: '#/channels/userSignedup/messages/UserSignedUp'
    x-recursion: # this RECURSION clips correctly
      $ref: '#'
components:
  messages:
    UserSignedUp:
      x-recursion: #this RECURSION clips correctly
        $ref: '#'
      payload:
        x-recursion: # HERE  <------------------------
          $ref: '#'
        type: object
        properties:
          x-recursion: #this RECURSION clips correctly
            $ref: '#'
          displayName:
            type: string
            description: Name of the user
          email:
            type: string
```

I found all the recursions being handled correctly.

except for the corner case recursion at `HERE  <------------------------`

this recursion is making the browser tab hang and go on in an infinite loop. weird .

will look into its root cause and update the PR with a fix accordingly

reachaadrika:
> > Add yes adding Error Boundary , Would be beneficial in the long run . Let's keep this isolated you can create a new issue for the same as Improvement .
> 
> Perfect 👍
> 
> > Let's test with 1-2 samples
> 
> while testing with the following yaml
> 
> ```yaml
> asyncapi: 3.0.0
> info:
>   title: Account Service
>   version: 1.0.0
>   description: This service is in charge of processing user signups
> channels:
>   userSignedup:
>     address: user/signedup
>     messages:
>       UserSignedUp:
>         $ref: '#/components/messages/UserSignedUp'
>     x-recursion:  #this RECURSION clips correctly
>       $ref: '#'
> operations:
>   sendUserSignedup:
>     action: send
>     channel:
>       $ref: '#/channels/userSignedup'
>     messages:
>       - $ref: '#/channels/userSignedup/messages/UserSignedUp'
>     x-recursion: # this RECURSION clips correctly
>       $ref: '#'
> components:
>   messages:
>     UserSignedUp:
>       x-recursion: #this RECURSION clips correctly
>         $ref: '#'
>       payload:
>         x-recursion: # HERE  <------------------------
>           $ref: '#'
>         type: object
>         properties:
>           x-recursion: #this RECURSION clips correctly
>             $ref: '#'
>           displayName:
>             type: string
>             description: Name of the user
>           email:
>             type: string
> ```
> 
> I found all the recursions being handled correctly.
> 
> except for the corner case recursion at `HERE <------------------------`
> 
> this recursion is making the browser tab hang and go on in an infinite loop. weird .
> 
> will look into its root cause and update the PR with a fix accordingly

we can check this case once , Let me know if you require any help here

catosaurusrex2003:
The issue arises due to a cyclic dependency between the Schema and Extensions components, as observed in the following code:
The code which is causing this issue is 
https://github.com/asyncapi/asyncapi-react/blob/076f36bd98cb867c26e3c8c2a5426b5551c1af4d/library/src/components/Schema.tsx#L361-L363
and

https://github.com/asyncapi/asyncapi-react/blob/076f36bd98cb867c26e3c8c2a5426b5551c1af4d/library/src/components/Extensions.tsx#L29-L31

But this cyclic dependency is required for the components to render correctly.

However, it's a rare edge case and might not need a fix. Instead we can choose to leave it as-is for now to avoid overcomplicating the solution.

Cases like these reinforce the need of implementing an Error Boundary. This would allow us to fallback to a previous, correct version of the AsyncAPI document (similar to how undo works in the editor with Ctrl + Z).

reachaadrika:
Agreed , Let's implement ErrorBoundary here . 
@AceTheCreator  LGTM , your reviews  ?

catosaurusrex2003:
I think this should be merged after https://github.com/asyncapi/asyncapi-react/pull/1108 is merged

AceTheCreator (approved):
LGTM!

AceTheCreator:
I think we can move this PR forward now @reachaadrika ?

PR #1108, opened 2024-11-13
feat: error boundary in layout to handle application errors / crashes

catosaurusrex2003:
**Description**
As proposed and discussed here https://github.com/asyncapi/asyncapi-react/pull/1101#issuecomment-2462910610.

1. Added an [error boundary](https://github.com/bvaughn/react-error-boundary) in layout to handle errors generated by infinite recursion and such so that the application doesnt crash into an irrecoverable state.

2. Reused the already existing Error component to render error,

**Related issue(s)**
https://github.com/asyncapi/asyncapi-react/issues/1100

### DIFF
--- a/library/src/containers/AsyncApi/Layout.tsx
+++ b/library/src/containers/AsyncApi/Layout.tsx
@@ -13,6 +13,7 @@ import { Error } from '../Error/Error';
 import { ConfigInterface } from '../../config';
 import { SpecificationContext, ConfigContext } from '../../contexts';
 import { ErrorObject } from '../../types';
+import AsyncApiErrorBoundary from '../ApplicationErrorHandler/ErrorBoundary';
 
 interface Props {
   asyncapi: AsyncAPIDocumentInterface;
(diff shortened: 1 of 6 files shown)


catosaurusrex2003:
@reachaadrika could you please review this pull request too when you get a chance? 😄 Thanks in advance!

library/src/containers/AsyncApi/Layout.tsx
```diff
@@ -48,24 +49,26 @@ const AsyncApiLayout: React.FunctionComponent<Props> = ({
...
+          <AsyncApiErrorBoundary>
+            <div
+              className={`${observerClassName} relative md:flex bg-white leading-normal`}
+              id={config.schemaID ?? undefined}
+              ref={ref}
+            >
+              {configShow.sidebar && <Sidebar />}
+              <div className="panel--center relative py-8 flex-1">
+                <div className="relative z-10">
+                  {configShow.errors && error && <Error error={error} />}
```
AceTheCreator, line 61: Is this line still relevant? i mean since you've imported the error boundary in the layout container and it already includes the error component.

AceTheCreator (commented):
awesome improvement, but i noticed when i fix the recursive properties in the schema, the error still persists in the UI until i either refresh or re-paste a valid schema 🤔

library/src/containers/AsyncApi/Layout.tsx
AceTheCreator: > This line handles the errors in our document when parsing it. The errors are in turn diagnostics with severity 0.

Aren't we already handling these errors here https://github.com/asyncapi/asyncapi-react/blob/afd5e8c902073229e56a8ca98c9a6967724685f6/library/src/containers/AsyncApi/Standalone.tsx#L79 ?

AceTheCreator: > The error boundary, on the other hand, is used to catch unexpected application errors, such as a recursion error, and is a separate concern.

I agree with you 👍🏾

my question now is if `{configShow.errors && error && <Error error={error} />}` is relevant if we are already handling it in the standalone code, don't you think?

catosaurusrex2003: nice observation 👀

As far as i can tell, here in [Standalone](https://github.com/asyncapi/asyncapi-react/blob/afd5e8c902073229e56a8ca98c9a6967724685f6/library/src/containers/AsyncApi/Standalone.tsx#L74-L85). Whenever there is an error in the document. Our `asyncapi` variable becomes undefined. So the if else statements here in standalone catch the error.

And it never reaches our Error component in [Layout](https://github.com/asyncapi/asyncapi-react/blob/afd5e8c902073229e56a8ca98c9a6967724685f6/library/src/containers/AsyncApi/Layout.tsx#L59C1-L59C72).

So maybe we can remove the Error component rendering in Layout. Correct me if i am wrong or missing out on something.

Or we can move the conditional statements into Layout to check in one place only.

catosaurusrex2003: with some modification of prop types
we can put this in Layout.tsx
``` tsx
  return (
    <ConfigContext.Provider value={config}>
      <section className="aui-root">
        {asyncapi ? (
          <SpecificationContext.Provider value={asyncapi}>
            <div
              className={`${observerClassName} relative md:flex bg-white leading-normal`}
              id={config.schemaID ?? undefined}
              ref={ref}
            >
              {configShow.sidebar && <Sidebar />}
              <div className="panel--center relative py-8 flex-1">
                <div className="relative z-10">
                  {configShow.info && <Info />}
                  {configShow.servers && <Servers />}
                  {configShow.operations && <Operations />}
                  {configShow.messages && <Messages />}
                  {configShow.schemas && <Schemas />}
                </div>
                <div className="panel--right absolute top-0 right-0 h-full bg-gray-800" />
              </div>
            </div>
          </SpecificationContext.Provider>
        ) : (
          configShow.errors && error && <Error error={error} />
        )}
      </section>
    </ConfigContext.Provider>
  );
```

which yeilds the same effect.


Let me know your thoughts on this.

catosaurusrex2003:
> but i noticed when i fix the recursive properties in the schema, the error still persists in the UI until i either refresh or re-paste a valid schema 🤔

Thanks for highlighting that. Have put some logic to rerender the component fresh.

library/src/containers/AsyncApi/Layout.tsx
catosaurusrex2003: My reasoning behind moving the `Error` component to layout was that all the component rendering logic should be in one place, that is Layout.

But moving it to standalone will also do the trick, will fix that :+1: .

AceTheCreator (approved):
LGTM!

Issue #1109, opened 2024-11-25
Sidebar: wrong anchor hash for messages


FelixPopp:
**Description**
The anchor link for a message in the sidebar does not work as expected when `message.name()` and `message.id()` differ.

**Expected result**
Clicking on a message in the sidebar should set the correct hash in the URL and scroll the message into view. 

**Actual result**
The anchor link only works if `message.name()` and `message.id()` are identical.
If `message.name()` is not provided, the hash becomes `#message-undefined`.

**Steps to reproduce**
1. [Open the AsyncAPI React Playground App](https://asyncapi.github.io/asyncapi-react)
2. Enable the sidebar in the configuration tab
3. Links in the sidebar for message `lightMeasured Message ID` and `LwM2mOjbects` won't work. 
Example: The anchor hash for `LwM2mOjbects` is `#message-undefined` because `message.name` is not provided.
![Screenshot 2024-11-25 122934](https://github.com/user-attachments/assets/d93d069d-cf6f-4c27-ba28-a07d5d299743)

**Troubleshooting**
- The HTML-ID for a message is set to `message.id()` in [Messages.tsx#L32](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/containers/Messages/Messages.tsx#L32)
- However, the anchor links in the sidebar use `message.name()` as the hash target, defined in [Sidebar.tsx#L44](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/containers/Sidebar/Sidebar.tsx#L44)

parthikjadav:
ya i chacked it and thats true it is having error but i had fixed it

PR #1116, opened 2024-12-03
fix(sidebar): resolved sidebar hash URL navigation issue

Aditya0732:
fixed this issue #1109 

Description

The anchor link for a message in the sidebar does not work as expected when message.name() and message.id() differ.

Expected result
Clicking on a message in the sidebar should set the correct hash in the URL and scroll the message into view.

Actual result
The anchor link only works if message.name() and message.id() are identical.
If message.name() is not provided, the hash becomes #message-undefined.

before
before

after
after

Code changes required : 
async-sidebar


AceTheCreator (approved):
LGTM!

PR #1139, opened 2025-03-11
fix: multischema schema level in message payload

### LINKED ISSUE #1125: No generated examples when using multi format schema object
**Description**

https://www.asyncapi.com/docs/reference/specification/v3.0.0#messageObject
> **payload**
> _Multi Format Schema Object | Schema Object | Reference Object_
> Definition of the message payload. If this is a Schema Object, then the schemaFormat will be assumed to be "application/vnd.aai.asyncapi+json;version=asyncapi" where the version is equal to the AsyncAPI Version String.

The AsyncAPI spec says we can provide the schemas in 3 ways:
1. [A inline schema](https://www.asyncapi.com/docs/reference/specification/v3.0.0#schemaObject)
1. [A reference to a schema](https://www.asyncapi.com/docs/reference/specification/v3.0.0#referenceObject)
1. [A object with the keys `schemaFormat` and `schema` where `schema` will be either a inline schema or a reference to a schema](https://www.asyncapi.com/docs/reference/specification/v3.0.0#multiFormatSchemaObject)

If you use option 3 then [MessageHelpers.generateExample](https://github.com/asyncapi/asyncapi-react/blob/master/library/src/helpers/message.ts#L12) will call the sample generator with the `{"schemaFormat": ..., "schema": ...}` root object instead of the actual schema that is found under `schema.schema`. 😅 

Image

It is possible to work around this in some cases by providing the schema using option 1 or 2 if you only use a syntax that is compatible with `application/vnd.aai.asyncapi+json` but that don't support AVRO based schemas for example.

**Fix:**
Check in https://github.com/asyncapi/asyncapi-react/blob/master/library/src/helpers/message.ts#L9 if the object matches the shape of a "multi format schema object" and call the ´schema.schema` in that case.

I can provide a PR if that fix is acceptable.

**Expected result**
A basic example

**Actual result**
A empty example

AceTheCreator:
This PR updates the handling of multi-format schemas by restructuring the schema property at the top level. 

This changes ensure compatibility with the sample generator by flattening the schema structure while preserving its key properties.

fixes #1125


catosaurusrex2003:
Also I think there should be tests for the `MessageHelpers.generateExample` for Multiformat schema.

AceTheCreator:
> ere should be tests for the `MessageHelpers.generateExample` for Multiformat schema.

I mean it's not just the 

> Also I think there should be tests for the `MessageHelpers.generateExample` for Multiformat schema.

Added ✅

library/src/helpers/message.ts
```diff
@@ -8,8 +8,16 @@ export class MessageHelpers {
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
   static generateExample(schema: any, options: any = {}) {
     try {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
catosaurusrex2003: Do you think there should be a comment here to explain that we are flattening only in case of multi-format schema cuz the conditionals are hard to read ?

Up to you.

PR #1176, opened 2025-09-18
Feat/1152 activity loader

### LINKED ISSUE #1152: Add Activity Loader on Initial Page Load
Right now, when the page loads for the first time, there’s a flash where it renders without styles, it basically looks like raw HTML for a moment. This creates a broken experience and makes it feel like something’s broken, especially on slower networks.

To smooth things out, we should add an activity loader that shows while the styles and components are rendering. This gives users some visual feedback and prevents the weird unstyled flash.

So here's what should happen:

- Add a basic loading indicator(I'll probably prefer a skeleton 💀 ) that show on initial load
- Ensure it disappears once everything is styled and rendered properly
- We should probably keep it simple and lightweight so users don't experience noticeable delay
- Ensure it is tested on slow 3g

utkarshgupta2804:
Summary of Changes
1. Added a new loader component for improved user experience during content loading.
2. Created a loader folder inside components.
3. Implemented skeletonComponents.tsx, PlaygroundSkeleton.tsx, and index.ts for reusable skeleton UI.
4. Updated page.tsx to integrate the new loader seamlessly.

Testing
1. Verified loader functionality across different network conditions.
2. Confirmed consistent rendering and smooth user experience.

Resolves #1152 

Preview
Demonstration of the loader in action:

https://github.com/user-attachments/assets/c1859e62-411f-4574-bf65-d0674ed7f443


AceTheCreator (commented):
Hey @utkarshgupta2804 👋🏽

Thanks for taking the time to work on this! A couple of quick notes to help for future contributions:

Before opening a PR for an issue, please indicate your interest in the issue so we can confirm it’s good to pick up.

In this case, the activity loader isn’t needed, the delay was coming from the playground itself, and since the playground isn’t actually used by end users, modifying it won’t have an impact.

I’ll go ahead and close this PR, but really appreciate the effort you put in. Please feel free to check out other open issues, we’d love to see more contributions from you.

Issue #1161, opened 2025-06-08
POC: Auto-scroll and highlight updated component in docs preview


catosaurusrex2003:
**Description**
This is a proof of concept to explore how we can implement this in our playground text editor and the documentation preview.

📌 Goal:
Whenever a user makes a change in the AsyncAPI spec via the text editor, the docs preview should:
- Auto-scroll to the component or section that was updated.
- Temporarily highlight the updated section to draw attention (example of desired effect: [GIF example](https://www.asmitbm.me/articles/asyncapi/notifications_gif.gif), just the blue box effect—UI style not relevant).

This makes it significantly easier for users to:
- Visualize the impact of their changes.
- Avoid the hassle of manually scrolling through large specs to locate the updated element.

Future Scope:
🔭 Once the core functionality is implemented, we can improve UX further:
- Recognize that a single change can affect multiple places in the docs.
- Implement a mini-map on the right side (inspired by IDEs) that shows the full structure of the docs.
- Use visual markers on the mini-map to indicate all updated sections, even if only one is scrolled-to and highlighted at a time.
- Like this ![Image](https://github.com/user-attachments/assets/933d2216-fd50-41ef-bad6-030757e696c6)

**Why this matters**
Right now, users must scroll and hunt through the docs to verify every change they make in the editor—this breaks the workflow and adds friction. A live scroll + highlight feature brings a smoother, more intuitive spec writing experience.

**Source/Inspiration**
https://github.com/asyncapi/asyncapi-react/issues/1152#issuecomment-2862792921

Sunil1532:
I am interested to work on this issue. please assign this issue to me

utkarshgupta2804:
Okay sure

AceTheCreator:
@catosaurusrex2003 I don't think we need this implementation... It's just a playground.

PR #1177, opened 2025-09-25
feat: implemented change tracker + highlighting

Paulie-Aditya:
**Description**

- When a user makes changes in the editor, the onChange handler triggers
- The ChangeTracker compares the old and new schemas
If changes are detected:
- The affected section is identified
- The Highlight component is rendered over that section
- The section is scrolled into view
- A blue highlight effect appears and fades out after 2 seconds

**Related issue(s)**
Resolves: #1161


AceTheCreator:
@catosaurusrex2003 I don't think we need this implementation... It's just a playground.

PR #1199, opened 2025-12-18
fix: do not omit patternProperties when properties is undefined

### LINKED ISSUE #1197: patternProperties is ignored when properties is absent
**Description**

The `SchemaProperties` components just returns `null` if `properties === undefined || !Object.keys(properties)`, that is, there is no element in `properties`. This affects every ui that uses asyncapi-react to render the document visually. For instance, here's a related issue https://github.com/asyncapi/studio/issues/1223. 

**Expected result**

`patternProperties` should be rendered regardless of whether there is an element in `properties` or not.

**Actual result**

`patternProperties` are not rendered when there is nothing to show in `properties`.

**Steps to reproduce**

use any ui that uses `asyncapi-react`, for example studio.asyncapi.com. 
paste this schema in the editor:
```
asyncapi: 3.0.0
info:
  title: Account Service
  version: 1.0.0
components:
  schemas:
    Shadow:
      type: object
      properties:
        id:
          type: string
          format: uuid
      patternProperties:
        "^[a-zA-Z0-9]+$":
            type: string

    Camera:
      type: object
      patternProperties:
        "^[a-zA-Z0-9]+$":
            type: string
```

You will see that `patternProperties` are rendered for `Shadow` but not for `Camera`.

theAnuragMishra:
**Description**

-  the current implementation of SchemaProperties component skips and returns null if `properties` is `undefined` which shows no `patternProperties` even if they are present
- this pr converts `properties` and `patternProperties` to empty objects in case they are undefined. If there's nothing to show, `Object.entries` maps nothing and nothing is rendered which is desired.
- this fixes the bug where `patternProperties` is omitted when `properties` is `undefined`.

**Related issue(s)**
closes #1197

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2466,10 +2466,11 @@
       }
     },
     "node_modules/@istanbuljs/load-nyc-config/node_modules/js-yaml": {
-      "version": "3.14.1",
-      "resolved": "https://registry.npmjs.org/js-yaml/-/js-yaml-3.14.1.tgz",
-      "integrity": "sha512-okMH7OXXJ7YrN9Ok3/SXrnu4iX9yOk+25nqX4imS2npuvTYDmo/QEZoqwZkYaIDk3jVvBOTOIEgEhaLOynBS9g==",
+      "version": "3.14.2",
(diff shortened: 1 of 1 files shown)


theAnuragMishra:
@AceTheCreator  have a look :)

PR #1208, opened 2026-01-28
feat: implemented asyncapi-social-plugin in playground

catosaurusrex2003:
**Description**
Implemented the https://github.com/AceTheCreator/asyncapi-social-plugin in playground.
Since the AsyncApiPlugin interface was introduced, it needs to be implemented in the playground.


catosaurusrex2003:
should be merged after 
https://github.com/AceTheCreator/asyncapi-social-plugin/pull/2 is merged

AceTheCreator:
Hey @catosaurusrex2003, can you resolve the conflicts in this PR? 🙏🏽

AceTheCreator (changes requested):
@catosaurusrex2003 just a quick review... If we are using an extension plugin, then I think we should disable the built-in `x-` for the info section. wyt?

catosaurusrex2003:
@AceTheCreator Just to confirm, do you mean to remove the inbuilt code for rendering the social media plugin altogether ?

AceTheCreator:
> @AceTheCreator Just to confirm, do you mean to remove the inbuilt code for rendering the social media plugin altogether ?

@catosaurusrex2003 not necessarily. I think disabling it from the default config would do 🤷🏽‍♂️

catosaurusrex2003:
let me rephrase it
Even if i remove the inbuilt XExtension component from the default config
the Info Extension section still renders like this
image
Should we then implement toggling of whether to show Info Extension or not in https://github.com/asyncapi/asyncapi-react/blob/master/docs/configuration/config-modification.md ?

Am i missing something ?

AceTheCreator:
Looking at this from another perspective, since the extension config is global and doesn’t target any specific extension, I’d rather any changes happen in the playground. Extensions are first-class citizens in schemas, so introducing specific dynamic behavior here doesn’t feel right.

Given that the plugin integration is already self-explanatory, I’d prefer we close this PR 😅

catosaurusrex2003:
i agree

PR #1215, opened 2026-02-02
fix(playground): resolve dev server crash caused by Next.js 16 upgrade

mishop-15:
**Context**
The `playground` package was recently upgraded from Next.js 14 to 16.1.5 (via Dependabot ~2 days ago).

**The Issue**
Next.js 16 enables Turbopack by default for development. However, this project uses a custom webpack configuration in `next.config.mjs`. This causes `npm start` to fail immediately with a conflict error because Next.js cannot use both Turbopack and a custom Webpack config simultaneously without an explicit flag.

**The Fix**
1. Updated the `start` script in `playground/package.json` to add the `--webpack` flag. This forces Next.js to use the existing webpack config.
2. Removed the deprecated `eslint` key from `next.config.mjs` (which is no longer supported in Next.js 16 and was throwing warnings).
3. Added `next.config.mjs` to the `include` list in `playground/tsconfig.json` to resolve local ESLint parsing errors.

**Verification**
Validated that `npm start` now successfully launches the playground on `http://localhost:3000`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,7 @@ The list of props for the AsyncAPI React component includes:
 
 - **schema: string | AsyncAPIDocument | object | FetchingSchemaInterface**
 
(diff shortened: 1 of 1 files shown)


mishop-15:
> Please mention issue number in the PR description

There isn't an issue associated issue raised with this. I encountered while setting up the project locally.

Issue #1236, opened 2026-02-15
Error: "Something went wrong too much recursion. Please check document for recursion" for the document without recursion


lukaszzazulak:
**Description**

Document does not contain recursion, but documentation component says:
Error: "Something went wrong too much recursion. Please check document for recursion" 

Go to https://studio.asyncapi.com/ and paste the following content:
```
asyncapi: "2.6.0"
channels:
  queue:
    publish:
      x-consumers:
        c1:
          - $ref: "https://gist.githubusercontent.com/lukaszzazulak/23189f78612fdeabe1534fc0607f8055/raw/4254b7bc8b7b2f29acc10f8a320293b2652b14c9/gistfile1.txt#/components/x-principals/ecp-queue.sit-user"
        c2:
          - $ref: "https://gist.githubusercontent.com/lukaszzazulak/23189f78612fdeabe1534fc0607f8055/raw/4254b7bc8b7b2f29acc10f8a320293b2652b14c9/gistfile1.txt#/components/x-principals/ecp-queue.sit-user"
info:
  title: "Queue"
  version: "1.0.0"

```

**Expected result**

No error in the documentation component

**Actual result**

Error: Something went wrong
too much recursion. Please check document for recursion.

**Steps to reproduce**

See description

AceTheCreator:
@lukaszzazulak i guess we can close this issue now 🙏🏽

PR #1241, opened 2026-02-18
fix: recursion detection to allow same resolved $ref in different branches

catosaurusrex2003:
## The Problem

The recursion detection algorithm incorrectly throws "too much recursion" errors when the same external `$ref` appears in multiple branches (e.g. in the schema provided in #1236  `c1` and `c2` both referencing the same external component). 

This happens because the algorithm used a global `WeakSet` that flagged any object seen before, even if it was in a different branch.

## Solution

Replaced global `WeakSet` tracking with path-based tracking using `Set<object>`. Each branch now gets its own copy of the path, allowing the same object to appear in different branches while still detecting true circular references.

### Key Changes

- Changed `WeakSet` → `Set<object>` to enable path copying
- Renamed `visited` → `visitedInPath` for clarity
- Create new path copy for each branch instead of sharing global set

**Related issue(s)**

#1236


MichakrawSB:
@magicmatatjahu @fmvilas @AceTheCreator any updates?

MichakrawSB (approved):

YOU54F:
Would also like to see this merged, @catosaurusrex2003 are you able to rebase as there are some conflicts with the base branch.

Are the changes in the playground necessary (tsconfig / package.json) for this change, or were they just for testing purposes?

I would probably advise squashing into a single commit and use conventional commit notation in your message `fix: false positive recursion error when same external appears in multiple branches` to match the existing convention in the repo

AceTheCreator (commented):
Hey @catosaurusrex2003, just a small perf improvement. 

I like the idea of creating a new path copy for each branch instead of sharing a global set. What if, during the recursion, we clean up after? currently the clone copies the entire Set at every level. For every tree N levels deep, that's O(N²) work just for tracking.

catosaurusrex2003:
Nice suggestion
I will look into this and update the pr asap

catosaurusrex2003:
Thanks for the suggestion @AceTheCreator
I implemented the performance improvement.

It no longer creates a new set on ever level.
Current implementation is adding and deleting in the same set.

AceTheCreator (approved):
LGTM! 💜

Torgh:
Hi, could you please let me know when the merged changes are expected to be delivered? This delivery is very important to us.

PR #1272, opened 2026-04-27
fix: cleans up plugin event listeners on update and unmount properly

### LINKED ISSUE #1235: Event listeners are not properly cleaned up in AsyncApiComponent
**Description**

Event listeners registered for plugin events are not reliably removed during component updates because different function references are passed to `pm.on()` and `pm.off()`.

This causes listeners to accumulate over time, resulting in duplicated event emissions and potential memory leaks when the component re-renders or updates.

**Expected result**

Event listeners should be removed using the exact same handler reference that was used during registration, ensuring:

- no duplicated events
- proper cleanup during updates and unmount
- stable listener lifecycle

**Actual result**
```
 private setupEventListeners() {
    const { onPluginEvent } = this.props;
    const { pm } = this.state;

    if (!onPluginEvent) return;

    PLUGINEVENTS.forEach((event) => {
      pm?.on(event, this.handler(event));
    });
  }

  private cleanupEventListeners() {
    const { pm } = this.state;
    PLUGINEVENTS.forEach((event) => {
      pm?.off(event, this.handler(event));
    });
  }
```
Right now the Listeners are registered and cleaned up using:
```
pm.on(event, this.handler(event));
pm.off(event, this.handler(event));
```
Since `this.handler(event)` creates a new function instance on every call, `off()` does not receive the same reference that was originally registered with `on()`.

**Proposed solution**
Maintain stable handler references for each plugin event.
- Create handlers once (for example in the constructor).
- Store them in a `Map<string, Function>` keyed by event name.
- Reuse the stored handler for both `pm.on()` and `pm.off()`.
```
private eventHandlers = new Map<string, (data: unknown) => void>();

PLUGINEVENTS.forEach((event) => {
  this.eventHandlers.set(event, (data) => {
    this.props.onPluginEvent?.(event, data);
  });
});
```
and then,
```
pm.on(event, handler);
pm.off(event, handler);
```
**Steps to reproduce**
1. Create a plugin that periodically emits a plugin event.
2. Render `AsyncApiComponentWP` with that plugin and an `onPluginEvent` callback.
3. Change the `onPluginEvent` handler identity (for example, by updating state via a button click).
4. Observe the console output (each emission logged many times)

**Additional context**
- This issue is not limited to `onPluginEvent` identity changes. It can also occur when listener setup/cleanup runs with new handler references or when the plugin manager instance changes, leaving listeners attached to the previous instance.
- Listeners should be rebound if the plugin manager instance changes (`prevState.pm !== pm`) to avoid leaving listeners attached to an old instance.

Happy to open a PR for this if the maintainers approve.

Ishita-190:
**Description**

- created a handler function for each event, store it in the `eventHandlers` Map which will then be reused for both `pm.on` and `pm.off`
- Added `componentWillUnmount` to `AsyncApiComponent` for proper event listener cleanup.

verified this using a test, it's working properly :

Screenshot 2026-04-27 002536

**Related issue(s)**

Fixes #1235


AceTheCreator (changes requested):
Hey @Ishita-190 

Can you fix your failing tests? Once done, i'll give it a review

library/src/containers/AsyncApi/Standalone.tsx
```diff
@@ -26,6 +26,10 @@ interface AsyncAPIState {
 }
 
 class AsyncApiComponent extends Component<AsyncApiProps, AsyncAPIState> {
+  componentWillUnmount() {
```
AceTheCreator: I find the lifecycle placement hard to read. I don't think `componentWillUnmount` should be at the top-level

library/src/containers/AsyncApi/Standalone.tsx
```diff
@@ -125,26 +129,27 @@ class AsyncApiComponent extends Component<AsyncApiProps, AsyncAPIState> {
     );
   }
 
-  private handler(eventName: string) {
-    return (data: unknown) => {
-      this.props.onPluginEvent!(eventName, data);
-    };
-  }
+  private eventHandlers = new Map<string, (data: unknown) => void>();
```
AceTheCreator: Make this `readonly`

AceTheCreator (changes requested):
Left you a review :)

library/src/containers/AsyncApi/Standalone.tsx
Ishita-190: Sir, I've placed it below `componentDidUpdate`, I hope this fixes the issue.

library/src/containers/AsyncApi/Standalone.tsx
Ishita-190: done

catosaurusrex2003:
@Ishita-190 Oops, I just realized my PR might overlap or conflict with what this PR is doing:

https://github.com/asyncapi/asyncapi-react/pull/1296

@AceTheCreator, what do you think would be the best way to handle this? Should we merge one approach into the other, or close one of the PRs?

The only thing missing in my PR is code specific to `componentWillUnmount` function.

Ishita-190:
@catosaurusrex2003 oh no problem! whatever works is fine by me, `componentWillUnmount` is just for cleanup so maybe you can add it to your pr and I'll close this one.

AceTheCreator:
Yea, @catosaurusrex2003 i agree with @Ishita-190 ✅

catosaurusrex2003:
ok I will add the functionality specific to `componentWillUnmount` in my pr

sorry for the overlap @Ishita-190

Ishita-190:
@catosaurusrex2003 no problem! Closing this now

PR #1277, opened 2026-05-02
fix: add node_auth for npm release fallback

AceTheCreator:
Add a fallback for failing trusted publishing on the web-component release workflow.

cc @catosaurusrex2003


catosaurusrex2003:
Why is trusted publishing failing tho ?

AceTheCreator:
> Why is trusted publishing failing tho?

I believe it needs to be enabled in the package manager for the npm package before it can work. I’ve reached out to Lukasz to confirm. Meanwhile, we can fall back to using an environment variable.

catosaurusrex2003 (approved):

PR #1185, opened 2025-11-12
feat: add plugin support for the react component

AceTheCreator:
This PR introduces a plugin system for the AsyncAPI React component, enabling developers to extend and customize the documentation renderer with custom components, event handling, and dynamic plugin management.

### Changes

- New AsyncApiPlugin interface and PluginAPI for plugin development
- Component registration in PluginSlot.OPERATION slots
- Event system: on(), off(), emit() for inter-plugin communication
- Dynamic plugin registration/unregistration at runtime
- Priority-based component ordering (higher priority renders first)
- Full TypeScript support

```
const plugin: AsyncApiPlugin = {
  name: 'my-plugin',
  version: '1.0.0',
  install(api: PluginAPI) {
    api.registerComponent(PluginSlot.OPERATION, MyComponent, { priority: 150 });
    api.on('event-name', (data) => console.log(data));
  }
};

<AsyncApi schema={schema} plugins={[plugin]} />
```
Fixes #914

### Breaking Changes
None

PR #1296, opened 2026-06-05
fix: improve plugin error handling and lifecycle events

### LINKED ISSUE #1239: Improve plugin install/init error handling
**SUMMARY**

Make plugin installation and event handling safer by isolating plugin failures and using one consistent plugin:error payload across the runtime.

**SCOPE**

- Plugin registration/install failure isolation

- Event listener failure isolation during dispatch

- Standardized plugin:error payload  (additive, non-breaking)

- Replace index-based plugin event usage with explicit constants

- Tests for all above runtime error paths

- Integration-layer alignment to avoid inconsistent/duplicate error emission

**TASKS**

 **Add failure-first tests for plugin runtime behavior**

-  Add test: plugin install() throws, manager still works

-  Add test: failed plugin is not stored

- Add test: healthy plugin still registers after a failed one

-  Add test: plugin:error is emitted with expected fields

- Add test: if one listener throws during emit(), other listeners still run

 **Define canonical PluginErrorPayload (additive type)**

- include: source, phase, pluginName, message, cause, recoverable, timestamp (+ optional eventName)

**Replace index-based plugin error event usage**

- add explicit plugin event constants

- remove runtime dependence on PLUGINEVENTS[1]

**Harden PluginManager error boundaries**

- wrap plugin.install(api) in try/catch

- emit canonical plugin:error on install failure

- register plugin only after successful install

- guard each listener invocation during emit() and continue dispatch

**Align Standalone.tsx error emission**

- avoid duplicate plugin:error emission for the same failure path

I'm happy to open a PR for this issue.

catosaurusrex2003:
### **Description**

This PR hardens the plugin system which was implemented in https://github.com/asyncapi/asyncapi-react/pull/1185 so registration failures are handled consistently, lifecycle events are typed and documented, and `onSpecLoaded` behaves predictably when a spec is already present.

- **`PluginManager.register()` now returns `boolean`** — `true` on success, `false` if the plugin is already registered or `install()` throws. 

- Install failures are emitted as `plugin:error` with a structured `PluginErrorPayload` and isolate event listener failures during `emit()` so one bad callback does not stop others.
```ts
export interface PluginErrorPayload {
  pluginName: string;
  message: string;
  timestamp: number;
}

```

- Lifecycle event constants are exported as`PLUGIN_EVENT_READY`, `PLUGIN_EVENT_ERROR`, and `PLUGIN_EVENT_SPEC_LOADED` .

-  `plugin:ready` and `plugin:error` are forwarded to `onPluginEvent`; `specLoaded` is internal to `api.onSpecLoaded()`.

- **`onSpecLoaded` fires immediately when a schema is already loaded** — plugins registered after the spec is parsed receive the current schema right away; subsequent schema updates still trigger the callback via `updateContext`.

- **Resilient event dispatch** — a throwing listener (including `onPluginEvent`) is logged and skipped; remaining listeners still run. Listener failures are not re-emitted as `plugin:error` to avoid recursion.

- **Slot rendering fix** — which I have commented about here [here](https://github.com/asyncapi/asyncapi-react/pull/1185#discussion_r3358162162)

- Removed duplicate error emission from `Standalone` and fix plugin event listener cleanup

- Updated the doc to better reflect the events and features of the plugin system.

**Related issue(s)**
Fixes: https://github.com/asyncapi/asyncapi-react/issues/1239


docs/features/plugins.md
```diff
@@ -9,7 +9,14 @@ The AsyncAPI React component supports a flexible plugin system to extend and cus
...
+import {
+  AsyncApiPlugin,
+  PluginAPI,
+  PluginSlot,
+  PLUGIN_EVENT_ERROR,
+  PLUGIN_EVENT_READY,
+  PluginErrorPayload,
+} from '@asyncapi/react-component';
 
 const myPlugin: AsyncApiPlugin = {
```
AceTheCreator: Let's move `myPlugin` up a block so it can be shared between the import example and the dynamic registration example. That makes things more readable and avoids the confusion of `myPlugin` appearing undefined in the dynamic example.

playground/app/page.tsx
```diff
@@ -19,6 +26,31 @@ import * as specs from '@/specs';
 
 const defaultSchema = specs.streetlights;
 
+const PlaygroundTestPluginInfo = () => (
```
AceTheCreator: I don't think we should add the plugin implementation unless we have a functional one to include in the playground.

library/src/helpers/pluginManager.ts
```diff
@@ -19,24 +29,54 @@ class PluginManager implements MessageBus {
...
+    } catch (error) {
+      // Always log so failures are visible even without an `onPluginEvent` handler.
+      console.error(`Failed to register plugin ${plugin.name}:`, error);
+      this.emit(PLUGIN_EVENT_ERROR, {
+        pluginName: plugin.name,
+        message: error instanceof Error ? error.message : String(error),
+        timestamp: Date.now(),
+      });
+      return false;
+    }
```
AceTheCreator: This current implementation seems to only cover synchronous plugin installation. Async install not covered: `register()` calls `plugin.install(api)` synchronously; a plugin returning a rejected Promise would not be caught by current tests, and would create an unhandled rejection.

An example is this: 

```const asyncPlugin: AsyncApiPlugin = {
  name: 'async-plugin',
  version: '1.0.0',
  install: async (api) => {
    const result = await fetch('/plugin-data').then((res) => res.json());
    if (!result.ok) {
      throw new Error('plugin configuration failed');
    }
    api.registerComponent(PluginSlot.INFO, MyInfoComponent);
  },
};

```

I think we should make the register instance async:

```async register(plugin: AsyncApiPlugin): Promise<boolean> {
  ...
  try {
    await plugin.install(api);
  } catch (error) {
    ...
  }
  ...
}

```

And the plugin type should allow Promise returns:

```
install(api: PluginAPI): void | Promise<void>;
```

library/src/types.ts
```diff
@@ -110,6 +110,12 @@ export interface AsyncApiPlugin {
   install(api: PluginAPI): void;
 }
 
+export interface PluginErrorPayload {
+  pluginName: string;
+  message: string;
+  timestamp: number;
```
AceTheCreator: Why is timestamp a `number` type, instead of a `Date` type?

AceTheCreator (changes requested):
Amazing stuff @catosaurusrex2003 🚀

Left you a review 🙌🏾

playground/app/page.tsx
catosaurusrex2003: Ah yes, I forgot to remove these changes.
thanks

docs/features/plugins.md
catosaurusrex2003: makes sense
done :+1:

library/src/types.ts
catosaurusrex2003: makes sense.
done :+1:

library/src/helpers/pluginManager.ts
catosaurusrex2003: Done :+1: 

I added support for asynchronous plugin installation.

catosaurusrex2003:
@AceTheCreator I did the requested changes,

Pls give it a review again.

catosaurusrex2003:
Since the PR now also includes a new feature of support for asynchronous plugin installs

Should I rename the PR from `fix:` to `feat:` and better title

AceTheCreator:
> Since the PR now also includes a new feature of support for asynchronous plugin installs
> 
> Should I rename the PR from `fix:` to `feat:` and better title

We can leave it as a fix

library/src/helpers/pluginManager.ts
```diff
@@ -19,43 +33,97 @@ class PluginManager implements MessageBus {
...
+    try {
+      await plugin.install(api);
+    } catch (error) {
+      // Always log so failures are visible even without an `onPluginEvent` handler.
+      console.error(`Failed to register plugin ${plugin.name}:`, error);
+      this.removePluginComponents(plugin.name);
+      this.emit(PLUGIN_EVENT_ERROR, {
+        pluginName: plugin.name,
+        message: error instanceof Error ? error.message : String(error),
+        timestamp: new Date(),
```
AceTheCreator: I think we should use `Date.now()` here instead of creating a `Date` object, since we're only interested in the timestamp. `Date.now()` returns the timestamp directly and avoids the overhead of instantiating a `Date`.

I think the same applies to timestamp presence in the rest of the code

AceTheCreator (changes requested):
Left you a final review. We are good to go after that

library/src/helpers/pluginManager.ts
catosaurusrex2003: `Date.now()` returns a compact (13 digits) epoch number, as it was before you asked to change it [here](https://github.com/asyncapi/asyncapi-react/pull/1296#discussion_r3375539742). The number is

> the number of milliseconds elapsed since the [epoch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date#the_epoch_timestamps_and_invalid_date), which is defined as the midnight at the beginning of January 1, 1970, UTC. [source](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now).

Do you mean a timestamp by doing `new Date().toISOString()` which will be an ISO string like `2026-06-12T14:30:00.123Z` ?

IMO, a number epoch is better than a timestamp, as it is easier to handle, faster and more used in telemetries.

AceTheCreator: Yeah, I was thinking of using `new Date().toISOString()`. My preference here isn't really about performance but about readability. I don't think this is a scenario where the performance difference would be significant enough to matter. In this case, making it easier to understand is a higher priority, tbh.

catosaurusrex2003: done

catosaurusrex2003:
@AceTheCreator I have changed timestamp to an ISO string.
Pls review it again.

library/src/types.ts
```diff
@@ -107,7 +107,13 @@ export interface AsyncApiPlugin {
...
   description?: string;
 
-  install(api: PluginAPI): void;
+  install(api: PluginAPI): void | Promise<void>;
+}
+
+export interface PluginErrorPayload {
+  pluginName: string;
+  message: string;
+  timestamp: string;
```
AceTheCreator, line 116: Change this to Date too, and we should be good to go @catosaurusrex2003.

AceTheCreator (changes requested):

library/src/types.ts
catosaurusrex2003: We are using ISO timestamps ([decided here](https://github.com/asyncapi/asyncapi-react/pull/1296#discussion_r3420006455)) 

Which is a `string`

not a `Date` object


now that you mention it
I notice that the doc in `plugins.md`
is still using 

`  timestamp: Date;
`

which needs to be changed to `string`

AceTheCreator (approved):
LGTM! ❤️

catosaurusrex2003:
This PR also solves the issue: https://github.com/asyncapi/asyncapi-react/issues/1235

aeworxet:
@asyncapi/microgrant_team

PR #1335, opened 2026-08-12
fix: upgrade Jest to v30 and unblock jsdom test suites

AceTheCreator:
**Description**

Bumping `@asyncapi/protobuf-schema-parser` to `^3.8.2` moved its `protobufjs`dependency from `^7.4.0` to `^8.7.1`. protobufjs v8 constructs a `TextDecoder` at module scope, and jsdom does not implement the Encoding API, so every test suite using the `@jest-environment jsdom` docblock failed to load with `ReferenceError: TextDecoder is not defined - 5 suites, 31 tests, none of which were running.

Rather than pin the parser back, this upgrades the test toolchain (Jest 26 wasreleased in 2020) and polyfills the missing globals.


catosaurusrex2003 (approved):

catosaurusrex2003:
lgtm :rocket:

PR #1342, opened 2026-08-28
fix: bump use-resize-observer to v10 for React 19 support

boyney123:
**Description**

- Bump `use-resize-observer` from `^9.1.0` to `^10.0.0` in `library/package.json` (lockfile updated minimally; the transitive `@juggle/resize-observer` dependency goes away).
- Switch `library/src/containers/AsyncApi/Layout.tsx` to the named `useResizeObserver` export, since v10 removed the default export.
- Update the `use-resize-observer` mock in `library/src/__tests__/index.test.tsx` to match the named export.

**Why**

`use-resize-observer@9` declares `peerDependencies.react: "16.8.0 - 18"`. When `@asyncapi/react-component` is installed into a project using React 19, npm's resolver cannot satisfy that peer range and loops indefinitely, alternating React 18 and React 19 at the tree root — `npm install` never completes. We hit this as a 10-minute CI timeout while upgrading EventCatalog to React 19. pnpm only emits a peer warning, so it is easy to miss in development.

`use-resize-observer@10` (released 2026-07-31) supports React 18 and 19 with an open-ended `react >=18.2` peer range, drops the `react-dom` peer and its runtime dependency, and keeps the same `useResizeObserver({ ref, onResize })` API used in `Layout.tsx`.

How the v10 breaking changes affect this library:

- Default export removed → handled by the import change in this PR.
- React floor moves to 18.2 → the library already develops against `react ^18.2.0`; the public peer range (`>=18.0.0`) is unchanged.
- `onResize` now reports a measured `0` as `0` instead of `undefined` → the existing `width === undefined` guard remains safe; a `0` width falls through to the `container:xl` branch as before.
- `/polyfilled` entry point removed → not used here.

**Testing**

- `npm run test:unit` — 13 suites, 201 tests pass.
- `npm run build:prod` in `library` — ESM, CJS, UMD and standalone bundles build successfully (webpack 5 resolves the new `exports` map).
- `npm ci` from the updated lockfile installs cleanly; `npm ls use-resize-observer` reports a single `10.0.0`.

**Documentation**

No documentation changes needed — no public API or configuration changes.

**Related issue(s)**

See also event-catalog/eventcatalog#2828 (React 19 upgrade blocked by the `use-resize-observer@9` peer range).


AceTheCreator (approved):
LGTM!

boyney123:
Thanks @AceTheCreator !

